PR #5, opened 2025-11-14
[Fusilli] Add matmul node 2/4

IanWood1:
Add node and verification for matmul ops.

Old PR: https://github.com/nod-ai/shark-ai/pull/2632

### DIFF
--- a/include/fusilli/node/matmul_node.h
+++ b/include/fusilli/node/matmul_node.h
@@ -141,21 +141,6 @@ public:
               ", B has dim=" + std::to_string(bDimVal));
     }
 
-    // Check that batch dimensions are outermost and non-transposed.
-    // This is equivalent to checking that perm[i] == i for all batch dims.
-    auto checkBatchDims = [&](const std::shared_ptr<TensorAttr> &tensor,
-                              const std::string &name) {
-      std::vector<int64_t> perm = tensor->getLogicalToPhysicalPermuteOrder();
-      for (size_t i = 0; i < batchDims; ++i) {
-        FUSILLI_RETURN_ERROR_IF(
-            perm[i] != static_cast<int64_t>(i), ErrorCode::InvalidAttribute,
-            "Matmul input tensor " + name +
-                " has batch dimensions that are not outermost or are "
-                "transposed");
-      }
-      return ok();
-    };
-
     FUSILLI_CHECK_ERROR(checkBatchDims(aT, "A"));
     FUSILLI_CHECK_ERROR(checkBatchDims(bT, "B"));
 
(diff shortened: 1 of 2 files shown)


tests/test_matmul_node.cpp
```diff
@@ -0,0 +1,601 @@
...
+        TensorAttr().setDim({3, 4}).setStride({4, 1})));
+    attr.setC(std::make_shared<TensorAttr>(
+        TensorAttr().setDim({2, 4}).setStride({4, 1})));
+    MatmulNode node(std::move(attr), ctx);
+
+    FUSILLI_REQUIRE_OK(node.preValidateNode());
+  }
+}
+
+TEST_CASE("MatmulNode inferPropertiesNode when C is fully specified",
```
IanWood1: @a-sidorova from your comment on the old PR:
>Minor: if we want to validate the method  `inferPropertiesNode` in this and below test cases , can we leave only the call of `inferPropertiesNode`? As it's [done ](https://github.com/nod-ai/shark-ai/blob/acd1542fdb7e59cd20e6353e2ef5cf2a80f38079/sharkfuser/tests/test_conv_node.cpp#L126)for `convolution`

Would it be okay to rename this but keep the calls to `inferPropertiesNode` and `postValidateNode`? I think this is useful to check that the node fully validates and doesn't modify the output tensor.

include/fusilli/node/matmul_node.h
```diff
@@ -0,0 +1,204 @@
...
+        cRank < kNonBatchRank, ErrorCode::InvalidAttribute,
+        "Matmul output tensor C must have a rank of at least 2");
+
+    FUSILLI_RETURN_ERROR_IF(
+        cT->getDim() !=
+            getMatmulInferredOutputShape(aT->getDim(), bT->getDim()),
+        ErrorCode::InvalidAttribute,
+        "Matmul output tensor C dimensions do not match the expected shapes "
+        "inferred based on the input dimensions");
+    return ok();
```
a-sidorova: As far as I understand, there might be situation when user manually sets `cDim` and `cStride` (they are not inferred by `inferPropertiesNode`).

Probably we should validate `cStride` here too? For example, need to check that `cStride` has the rank `cRank`.

tests/test_matmul_node.cpp
a-sidorova: If we want to validate just `inferPropertiesNode`, I think we should call only `inferPropertiesNode` (or with other methods before this call in common pipeline).
If we would like to validate entire pipeline (check that the node fully validates), we need to the entire pipeline for sure.

I think since you are an author of these tests, you need to decide what we validate - so this is up to you :)

sjain-stanford (commented):
I think we need stride validations in pre/post for matmul specific constraints. Please ping me and I'll give it another look.

include/fusilli/node/matmul_node.h
```diff
@@ -0,0 +1,222 @@
...
+    // Output stride is contiguous (row-major) when unspecified.
+    if (cStride.empty()) {
+      cT->setStride(
+          generateStrideFromDim(cDim, getContiguousStrideOrder(cDim.size())));
+    }
+
+    return ok();
+  }
+
+  ErrorObject postValidateNode() const override final {
```
sjain-stanford: Can you replicate the `checkBatchDims` test for `cT` as well here? This will be useful for cases where users annotate the stride incorrectly and use transposes on batch dims.

sjain-stanford (approved):
LGTM modulo the `postValidateNode` to include output tensor stride checks.

sjain-stanford:
Please rebase on main to get the flake_tests going.

PR #21, opened 2025-11-20
[Benchmark Driver] Add flag to optionally dump compilation artifacts

sjain-stanford:
This should make debugging perf and sharing repros easier.


IanWood1 (approved):
LGTM! Not blocking this PR, but should we also reference `FUSILLI_CACHE_DIR` in the readme somewhere? Maybe this could be added to the "logging" section?

PR #22, opened 2025-11-20
[Build] Fetch catch2 if unavailable

sjain-stanford:
Motivation is to make the non-docker build a bit less inconvenient. 

Other minor changes include:
- Link libs for libutils that were previously missing
- Bump CLI11 git tag

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,19 @@ endif()
 
 # Both tests and benchmarks depend on libutils
 if(FUSILLI_BUILD_TESTS OR FUSILLI_BUILD_BENCHMARKS)
+  # Find prebuilt Catch2 library or fetch it if not found
+  find_package(Catch2 3 QUIET)
+  if(NOT Catch2_FOUND)
+    message(STATUS "Catch2 not found on system: Fetching from GitHub")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -151,14 +152,28 @@ if(FUSILLI_BUILD_TESTS OR FUSILLI_BUILD_BENCHMARKS)
   # Add library for tests/utils.h
   add_library(libutils INTERFACE)
   target_include_directories(libutils INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/tests)
+  target_link_libraries(libutils INTERFACE
+    libfusilli
+    Catch2::Catch2WithMain
+  )
```
AaronStGeorge: I would have assumed that using `Catch2::Catch2WithMain` before `find_package(Catch2 3)` (or equivalent) would cause an error. But I guess not 🤷‍♂️ .

sjain-stanford: Yeah it appears CMake doesn't immediately validate that link targets exist during the `target_link_libraries` call, probably defers target resolution until later. But this is fragile, let me refactor this a bit.

sjain-stanford: Done.

PR #31, opened 2025-12-04
[Build] Take a source dependency on IREE runtime

sjain-stanford:
This is required to enable integration into TheRock, since IREE runtime headers don't get "installed" which is how TheRock expects to bring CMake deps in.

### DIFF
--- a/build_tools/docker/exec_docker_ci.sh
+++ b/build_tools/docker/exec_docker_ci.sh
@@ -20,5 +20,5 @@ docker run --rm \
            -v "${PWD}":/workspace \
            ${DOCKER_RUN_DEVICE_OPTS} \
            --security-opt seccomp=unconfined \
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -44,23 +44,81 @@ list(APPEND CMAKE_MODULE_PATH
...
+#
+################################################################################
+
+# Version pins for dependencies.
+# Prefer to keep the IREE_GIT_TAG synced with the python-installed IREE version.
+# At a minimum, the compiler from those packages must be compatible with the
+# runtime at this source ref.
+set(IREE_GIT_TAG "iree-3.10.0rc20251201")
+set(IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
+
```
AaronStGeorge: I might put these variables inside the else clause of 
```
if(IREE_SOURCE_DIR)
  ...
else()
 ...
```

PR #39, opened 2025-12-09
[Build, HipdnnPlugin] Install plugin tests

AaronStGeorge:
TheRock CI generates `test`, `lib`, `dbg`, etc. artifacts from the build - CI tests are run on separate runners using a downloaded version of a project's test artifacts. This approach allows the build to use a cheaper CPU-only machine, and the tests for each project to run in parallel on more expensive GPU machines (after the build has completed).

For projects like fusilli plugin, the pattern in TheRock is to install test files as well as a generated CTest script in the `test` artifact. When the `test` artifact is  downloaded onto an isolated runner, CTest can simply be pointed at the generated script + pre-built tests. 

This PR adds `FusilliPluginTestUtils.cmake` + `add_fusilli_plugin_test()` to wrap up boilerplate for test setup + test installation + test script generation in the plugin.  Note all paths used in tests must be relative to support install relocation.

### DIFF
--- a/plugins/hipdnn-plugin/CMakeLists.txt
+++ b/plugins/hipdnn-plugin/CMakeLists.txt
@@ -99,6 +99,9 @@ target_compile_definitions(${FUSILLI_PLUGIN_NAME} PRIVATE
   FUSILLI_PLUGIN_NAME="${FUSILLI_PLUGIN_NAME}"
   FUSILLI_PLUGIN_ENGINE_ID=${FUSILLI_PLUGIN_ENGINE_ID}
 )
+target_link_libraries(
+    ${FUSILLI_PLUGIN_NAME} PRIVATE "-Xlinker --exclude-libs=ALL"
+) # Hide symbols from linked static libs
 set_target_properties(${FUSILLI_PLUGIN_NAME} PROPERTIES
   C_VISIBILITY_PRESET hidden
   CXX_VISIBILITY_PRESET hidden


sjain-stanford (approved):
I don't have a lot of feedback on how this interfaces with TheRock, and you might want to tag someone from hipdnn for feedback on how this "fits". Stamping to unblock.

plugins/hipdnn-plugin/CMakeLists.txt
```diff
@@ -18,12 +18,15 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
+include(GNUInstallDirs)
 include(FetchContent)
 
 # Global constants
 set(FUSILLI_PLUGIN_NAME fusilli_plugin)
 set(FUSILLI_PLUGIN_ENGINE_ID 1001)
-set(FUSILLI_PLUGIN_DIR "${CMAKE_CURRENT_BINARY_DIR}/lib/${HIPDNN_PLUGIN_ENGINE_SUBDIR}")
+# Relative path from build/install prefix to plugin location.
+# HIPDNN_PLUGIN_ENGINE_SUBDIR is provided by hipdnn_sdk.
+set(FUSILLI_PLUGIN_RELATIVE_PATH "${CMAKE_INSTALL_LIBDIR}/${HIPDNN_PLUGIN_ENGINE_SUBDIR}")
```
BrianHarrisonAMD: Don't you need a find_package(hipdnn_sdk) above this to be able to use this path? 

I am a bit confused how you are able to get the HIPDNN_PLUGIN_ENGINE_SUBDIR with it being included below.

plugins/hipdnn-plugin/test/integration/CMakeLists.txt
```diff
@@ -4,31 +4,22 @@
...
     test_convfprop.cpp
     test_basic.cpp
-)
-target_compile_options(fusilli_plugin_integration_tests PRIVATE ${HIPDNN_WARNING_COMPILE_OPTIONS})
-target_link_libraries(fusilli_plugin_integration_tests PRIVATE
+  DEPS
     GTest::gtest_main
     hip::host
     hipdnn_frontend
     hipdnn_sdk
```
BrianHarrisonAMD: Pretty soon your tests can use hipdnn_test_sdk which will bring in both these deps.
The [PR](https://github.com/ROCm/rocm-libraries/pull/3059) should be landing as soon as CI passes (was waiting on Submodule bumps).

plugins/hipdnn-plugin/test/integration/test_basic.cpp
```diff
@@ -47,11 +50,16 @@ static std::vector<std::string> getLoadedPlugins(hipdnnHandle_t handle) {
...
-  // Ensure hipDNN will load fusilli plugin.
-  const std::array<const char *, 1> paths = {FUSILLI_PLUGIN_DIR};
+  // Set plugin paths.
+  //
+  // FUSILLI_PLUGIN_PATH is a relative path from the executable directory where
+  // this test lives to the location of the plugin's .so e.g.
+  // "../lib/hipdnn_plugins/engines/fusilli_plugin". The tests will be
+  // installed, and therefore re-located, in some build configurations
+  // (`TheRock` for example) so we must use a relative path.
+  auto pluginPath = std::filesystem::canonical(getCurrentExecutableDirectory() /
```
BrianHarrisonAMD: I think we typically use weakly_canonical instead of canonical:

```
        // Note: The plugin paths has to be set before we create the hipdnn handle.
        auto pluginPath = std::filesystem::weakly_canonical(
            hipdnn_sdk::utilities::getCurrentExecutableDirectory() / PLUGIN_PATH);
```

I think we do this generally since the backend supports relative engine plugin loading, but I guess for tests you might not need this.

plugins/hipdnn-plugin/CMakeLists.txt
AaronStGeorge: Odd... Yeah that shouldn't work. It has been though ([fusilli plugin passing in TheRock](https://github.com/iree-org/fusilli/actions/workflows/build-and-test.yml)). Probably something to do with CMake defining things then actually evaluating them later or something else unholy. I'll fix that.

plugins/hipdnn-plugin/test/integration/test_basic.cpp
AaronStGeorge: For this test I think we want `canonical`, we know the plugin should be at that path, if it's not it would be better to fail.

plugins/hipdnn-plugin/CMakeLists.txt
BrianHarrisonAMD: Yea I hit the same thing actually.
It allows it, but seems like it's a quirk and not intended to be supported from the docs I read.

BrianHarrisonAMD (commented):
Not sure what support level the Fusilli plugin is at for Windows, but we hit some path issues between Linux / Windows with MIOpen plugin which you may want to test / check for your plugin as well.

If it works in TheRock with testing, and you can see that the install of TheRock has your plugin in the right location, then I think you are likely good to go.

plugins/hipdnn-plugin/CMakeLists.txt
AaronStGeorge: It looks like CMake uses empty string for undefined variables. So `FUSILLI_PLUGIN_RELATIVE_PATH` becomes `lib/` rather than `lib/hipdnn_plugins/engines/`. The former won't actually cause any issues in the tests we have, hence why this slipped through.

Issue #45, opened 2025-12-11
[Fusilli] Support Layernorm


a-sidorova:
TODO:

- [x] Attributes: https://github.com/iree-org/fusilli/pull/44
- [ ] Node: https://github.com/iree-org/fusilli/pull/46
- [ ] ASM Emitter: N/A
- [ ] Driver: N/A

PR #46, opened 2025-12-11
[Fusilli] Implemented Layernorm node

a-sidorova:
The second part of https://github.com/iree-org/fusilli/issues/45

### DIFF
--- a/include/fusilli/node/layernorm_node.h
+++ b/include/fusilli/node/layernorm_node.h
@@ -79,36 +79,47 @@ public:
                                 "defined by its stride");
 
     // Shape and layout checks on scale tensor.
+    // If scale tensor's dims/strides are not set, they will be inferred in
+    // inferPropertiesNode().
     if (sT) {
-      std::vector<int64_t> expectedDim = xT->getDim();
-      expectedDim[0] = 1;
-      FUSILLI_RETURN_ERROR_IF(sT->getDim() != expectedDim,
-                              ErrorCode::InvalidAttribute,
-                              "LayerNorm input tensor SCALE must have shape as "
-                              "tensor X with single batch");
+      if (!sT->getDim().empty()) {
+        FUSILLI_RETURN_ERROR_IF(
+            sT->getDim() != getScaleBiasDim(xT->getDim()),
+            ErrorCode::InvalidAttribute,
+            "LayerNorm input tensor SCALE must have shape as "
+            "tensor X with single batch");
+      }
 
-      FUSILLI_RETURN_ERROR_IF(
-          !sT->isContiguous() && !sT->isChannelsLast(),
-          ErrorCode::NotImplemented,
-          "Tensor '" + sT->getName() +
-              "' is neither contiguous nor channels-last as "
-              "defined by its stride");
+      if (!sT->getStride().empty()) {
+        FUSILLI_RETURN_ERROR_IF(
+            !sT->isContiguous() && !sT->isChannelsLast(),
+            ErrorCode::NotImplemented,
+            "Tensor '" + sT->getName() +
+                "' is neither contiguous nor channels-last as "
+                "defined by its stride");
+      }
     }
 
     // Shape and layout checks on bias tensor.
+    // If scale tensor's dims/strides are not set, they will be inferred in
+    // inferPropertiesNode().
     if (bT) {
-      std::vector<int64_t> expectedDim = xT->getDim();
-      expectedDim[0] = 1;
-      FUSILLI_RETURN_ERROR_IF(bT->getDim() != expectedDim,
-                              ErrorCode::InvalidAttribute,
-                              "LayerNorm input tensor BIAS must have shape as "
-                              "tensor X with single batch");
-      FUSILLI_RETURN_ERROR_IF(
-          !bT->isContiguous() && !bT->isChannelsLast(),
-          ErrorCode::NotImplemented,
-          "Tensor '" + bT->getName() +
-              "' is neither contiguous nor channels-last as "
-              "defined by its stride");
+      if (!bT->getDim().empty()) {
+        FUSILLI_RETURN_ERROR_IF(
+            bT->getDim() != getScaleBiasDim(xT->getDim()),
+            ErrorCode::InvalidAttribute,
+            "LayerNorm input tensor BIAS must have shape as "
+            "tensor X with single batch");
+      }
+
+      if (!bT->getStride().empty()) {
+        FUSILLI_RETURN_ERROR_IF(
+            !bT->isContiguous() && !bT->isChannelsLast(),
+            ErrorCode::NotImplemented,
+            "Tensor '" + bT->getName() +
+                "' is neither contiguous nor channels-last as "
+                "defined by its stride");
+      }
     }
 
     // Output tensor checks for training and inference forward phases.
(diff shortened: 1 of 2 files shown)


include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,263 @@
...
+#include <utility>
+#include <vector>
+
+namespace fusilli {
+
+//===----------------------------------------------------------------------===//
+// Layer normalization node.
+//===----------------------------------------------------------------------===//
+
+class LayernormNode : public NodeCRTP<LayernormNode> {
```
sjain-stanford: We also need Layernorm backward at some point (follow-on is fine). It's called `DLNNode` in cudnn: https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/include/cudnn_frontend/node/dln.h?L13

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,263 @@
...
+        layernormAttr.getForwardPhase() == NormFwdPhase::NOT_SET,
+        ErrorCode::AttributeNotSet, "Layernorm forward phase not set");
+
+    std::shared_ptr<TensorAttr> xT = layernormAttr.getX();
+    std::shared_ptr<TensorAttr> sT = layernormAttr.getSCALE();
+    std::shared_ptr<TensorAttr> bT = layernormAttr.getBIAS();
+    std::shared_ptr<TensorAttr> eT = layernormAttr.getEPSILON();
+    std::shared_ptr<TensorAttr> yT = layernormAttr.getY();
+    std::shared_ptr<TensorAttr> mT = layernormAttr.getMEAN();
+    std::shared_ptr<TensorAttr> vT = layernormAttr.getINV_VARIANCE();
```
IanWood1: These should be references, but I think this mistake is made throughout the codebase.

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,263 @@
...
+                              "batch dimension equal to "
+                              "input X tensor batch dimension");
+      // Stride check for output MEAN tensor
+      FUSILLI_RETURN_ERROR_IF(
+          stride != mT->getStride(), ErrorCode::InvalidAttribute,
+          "Layernorm output MEAN tensor must have unit strides");
+      // Stride check for output INV_VARIANCE tensor
+      FUSILLI_RETURN_ERROR_IF(
+          stride != vT->getStride(), ErrorCode::InvalidAttribute,
+          "Layernorm output INV_VARIANCE tensor must have unit strides");
```
IanWood1: Aren't all strides valid because these are both just scalars?

IanWood1 (approved):
LGTM!

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,263 @@
...
+    std::shared_ptr<TensorAttr> eT = layernormAttr.getEPSILON();
+    std::shared_ptr<TensorAttr> yT = layernormAttr.getY();
+    std::shared_ptr<TensorAttr> mT = layernormAttr.getMEAN();
+    std::shared_ptr<TensorAttr> vT = layernormAttr.getINV_VARIANCE();
+
+    // Ensure mandatory input and output tensors are set.
+    FUSILLI_RETURN_ERROR_IF(!xT, ErrorCode::AttributeNotSet,
+                            "Layernorm input tensor X not set");
+    FUSILLI_RETURN_ERROR_IF(!eT, ErrorCode::AttributeNotSet,
+                            "Layernorm input tensor EPSILON not set");
```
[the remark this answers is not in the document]
IanWood1: I was looking into this a bit more. [Here](https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/samples/cpp/norm/layernorm.cpp) is a sample in cudnn that uses it. I think its considered to be part of the attr vs being data like normal tensors. Both cudnn and Alexandra's PR https://github.com/iree-org/fusilli/pull/71 don't pass epsilon through the variant pack. It is a compile-time constant.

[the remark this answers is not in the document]
sjain-stanford: Thanks. 

> I'm just not sure if EPSILON should be in LayernormAttribute::InputNames.. But I decided to leave it so since cudnn-frontend has it [there](https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/include/cudnn_frontend/graph_properties.h?L1009).

I think we can drop `EPSILON` from the `InputNames` (since it is not really an op input) and just make it a regular private float on the `LayernormAttribute`. Then one could just create it from a float value instead of wrapping it with a tensor first. WDYT @IanWood1  @a-sidorova ?

```cpp
    auto layernormAttr = LayernormAttr()
                         .setEpsilon(1e-05f)
                            ...

```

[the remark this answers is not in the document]
a-sidorova: I was thinking about it. It's really simpler. Moreover, we shouldn't propagate it to variant pack (the [discussion](https://github.com/iree-org/fusilli/pull/71#discussion_r2682012512)) and `epsilon` is just compile-time scalar. One thing  stops me is that cudnn-frontend [implements ](https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/samples/cpp/norm/layernorm.cpp?L83)`epsilon` via tensor. If we wanna to work with just float value, there will be difference between fusilli and cudnn. But if it's ok for us, I will be glad to implement `epslion` using just float value

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,262 @@
...
+
+      std::shared_ptr<TensorAttr> mT = layernormAttr.getMEAN();
+      std::shared_ptr<TensorAttr> vT = layernormAttr.getINV_VARIANCE();
+
+      // Infer shape of output MEAN tensor.
+      if (mT->getDim().empty()) {
+        mT->setDim(dim);
+      }
+      // Infer shape of output INV_VARIANCE tensor.
+      if (vT->getDim().empty()) {
```
sjain-stanford: same

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,262 @@
...
+        vT->setDim(dim);
+      }
+      // Infer stride of output MEAN tensor.
+      if (mT->getStride().empty()) {
+        mT->setStride(stride);
+      }
+      // Infer stride of output INV_VARIANCE tensor.
+      if (vT->getStride().empty()) {
```
sjain-stanford: same

include/fusilli/node/layernorm_node.h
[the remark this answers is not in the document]
sjain-stanford: Thanks for pointing me to the `is_pass_by_value` discussion on your other PR. I added a comment there as well.

I'm fine with leaving this wrapped as a tensor and leaving it in the `InputNames` for now to lean into the consistency aspect.

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,256 @@
...
+                                "defined by its stride");
+
+    // Shape and layout checks on scale tensor.
+    if (sT) {
+      std::vector<int64_t> expectedDim = xT->getDim();
+      expectedDim[0] = 1;
+      FUSILLI_RETURN_ERROR_IF(sT->getDim() != expectedDim,
+                              ErrorCode::InvalidAttribute,
+                              "Layernorm input tensor SCALE must have shape as "
+                              "tensor X with single batch");
```
sjain-stanford: cudnn doesn't hard require dim/stride to be specified for scale and bias tensors (since they can be inferred based on "normalized shape" of input tensor). 

Perhaps you can keep this check if dim/stride are specified. And if not, then we want to infer them in `inferPropertiesNode` like in https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/include/cudnn_frontend/node/layernorm.h?L41-L56

sjain-stanford: Please also add unit tests for the different cases.

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,256 @@
...
+      std::shared_ptr<TensorAttr> mT = layernormAttr.getMEAN();
+      std::shared_ptr<TensorAttr> vT = layernormAttr.getINV_VARIANCE();
+
+      // Shape check for output MEAN tensor
+      FUSILLI_RETURN_ERROR_IF(
+          dim != mT->getDim(), ErrorCode::InvalidAttribute,
+          "Layernorm output MEAN tensor must have shape [B, 1, ..., 1] with "
+          "rank equal to shape rank of input X tensor and batch dimension "
+          "equal to "
+          "input X tensor batch dimension");
```
sjain-stanford: nit: remove unnecessary line splitting

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,256 @@
...
+      // Shape check for output INV_VARIANCE tensor
+      FUSILLI_RETURN_ERROR_IF(dim != vT->getDim(), ErrorCode::InvalidAttribute,
+                              "Layernorm output INV_VARIANCE tensor must have "
+                              "shape [B, 1, ..., 1] with "
+                              "rank equal to shape rank of input X tensor and "
+                              "batch dimension equal to "
+                              "input X tensor batch dimension");
```
sjain-stanford: same

include/fusilli/node/node.h
```diff
@@ -34,6 +34,7 @@ class INode {
     Pointwise,
     WGrad,
     DGrad,
+    Layernorm,
```
sjain-stanford:
```suggestion
    LayerNorm,
```

tests/test_layernorm_node.cpp
```diff
@@ -0,0 +1,471 @@
+// Copyright 2025 Advanced Micro Devices, Inc.
```
sjain-stanford:
```suggestion
// Copyright 2026 Advanced Micro Devices, Inc.
```

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,256 @@
...
+          vT, ErrorCode::InvalidAttribute,
+          "Layernorm output tensor INV_VARIANCE should not be set");
+    }
+
+    // Epsilon checks.
+    std::shared_ptr<TensorAttr> eT = layernormAttr.getEpsilon();
+    FUSILLI_RETURN_ERROR_IF(!eT, ErrorCode::AttributeNotSet,
+                            "Layernorm epsilon not set");
+    FUSILLI_RETURN_ERROR_IF(!eT->isScalar(), ErrorCode::InvalidAttribute,
+                            "Layernorm epsilon must be a constant scalar");
```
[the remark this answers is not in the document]
a-sidorova: Done, thanks!

include/fusilli/node/node.h
a-sidorova: Done, thanks!

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,256 @@
+// Copyright 2025 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
a-sidorova: Great catch, thanks!

include/fusilli/node/layernorm_node.h
a-sidorova: Oh, now I see why cudnn infers shape/stride of **input** tensors. Thank you for the idea!

Implemented with tests

include/fusilli/node/layernorm_node.h
```diff
@@ -0,0 +1,256 @@
...
+      FUSILLI_RETURN_ERROR_IF(
+          !sT->isContiguous() && !sT->isChannelsLast(),
+          ErrorCode::NotImplemented,
+          "Tensor '" + sT->getName() +
+              "' is neither contiguous nor channels-last as "
+              "defined by its stride");
+    }
+
+    // Shape and layout checks on bias tensor.
+    if (bT) {
```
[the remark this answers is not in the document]
a-sidorova: Done, thanks!

PR #50, opened 2025-12-12
[Fusilli][Graph] Added missed checks for non NULL tensors in node builders

a-sidorova:
The PR adds checks for non-nullptr tensors in node builders in `graph` when we populate names which are not set. It's needed to avoid segmentation fault when user may (accidentally or not) pass `nullptr` tensors as arguments of builders. 
We shouldn't throw exception here if tensor is `nullptr`, since the class `Node` is responsible for this validation (specially the method `preValidateNode()`, not builders.


a-sidorova:
> LGTM, please address the comment before landing.

Rebased and added missed checks in matmul node builders.
May I ask you to merge the PR when you have time please?

PR #55, opened 2025-12-15
Add matmul support to benchmark driver

IanWood1:
Adds matmul subcommand to the benchmark driver with the ability to control the M/N/K sizes, data type, and transposes on the RHS/LHS.

### DIFF
--- a/benchmarks/driver.cpp
+++ b/benchmarks/driver.cpp
@@ -181,8 +181,9 @@ static ErrorObject benchmarkMatmul(int64_t m, int64_t n, int64_t k, bool transA,
       transB ? std::vector<int64_t>{1, k} : std::vector<int64_t>{n, 1};
 
   Graph graph;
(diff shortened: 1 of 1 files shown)


benchmarks/driver.cpp
```diff
@@ -163,6 +163,79 @@ static ErrorObject benchmarkConvFprop(
...
+  // Build attributes based on transpose flags.
+  auto aDims = std::vector<int64_t>{m, k};
+  auto bDims = std::vector<int64_t>{k, n};
+  auto aStride =
+      transA ? std::vector<int64_t>{1, m} : std::vector<int64_t>{k, 1};
+  auto bStride =
+      transB ? std::vector<int64_t>{1, k} : std::vector<int64_t>{n, 1};
+
+  Graph graph;
+  auto graphName = std::format("benchmark_matmul_m{}_n{}_k{}_transA{}_transB{}",
```
sjain-stanford: Can you add dtype to the name here to disambiguate dumps of the same mnk/trans but diff dtypes?

PR #61, opened 2025-12-19
[NFC] Refactor benchmark driver

IanWood1:
- Creates two classes (`ConvOptions` and `MatmulOptions`) to separate the subcommand options and make it easier to pass them around.
- Extracts matmul/conv subcommand logic into functions (e.g. `registerConvOptions`).
- Extracts matmul/conv validation logic into functions (e.g. `runConvBenchmark`).
- Changes some error handling to use `ErrorObject` instead of integers.

### DIFF
--- a/benchmarks/driver.cpp
+++ b/benchmarks/driver.cpp
@@ -750,7 +750,7 @@ static int benchmark(int argc, char **argv) {
   if (convApp->parsed()) {
     ErrorObject status = runConvBenchmark(convOpts, iter, deviceId, dump);
     if (isError(status)) {
-      std::cerr << "Fusilli Benchmark failed: " << status << std::endl;
+      std::cerr << "Fusilli Conv Benchmark failed: " << status << std::endl;
       return 1;
     }
   }
@@ -758,7 +758,7 @@ static int benchmark(int argc, char **argv) {
   if (matmulApp->parsed()) {
     ErrorObject status = runMatmulBenchmark(matmulOpts, iter, deviceId, dump);
     if (isError(status)) {
-      std::cerr << "Fusilli Benchmark failed: " << status << std::endl;
+      std::cerr << "Fusilli Matmul Benchmark failed: " << status << std::endl;
       return 1;
     }
   }


benchmarks/driver.cpp
```diff
@@ -163,27 +193,28 @@ static ErrorObject benchmarkConvFprop(
   return ok();
 }
 
-static ErrorObject benchmarkMatmul(int64_t m, int64_t n, int64_t k, bool transA,
-                                   bool transB, int64_t iter, bool dump,
-                                   DataType matmulIOType, int64_t deviceId) {
+static ErrorObject benchmarkMatmul(const MatmulOptions &opts,
+                                   DataType matmulIOType, int64_t iter,
```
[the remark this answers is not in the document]
sjain-stanford: I think it's fine how it is now, since the flags (fp16, bf16 are part of opts and this is used to infer the final dtype).

benchmarks/driver.cpp
```diff
@@ -457,243 +492,271 @@ benchmarkConvDGrad(int64_t n, int64_t c, int64_t d, int64_t h, int64_t w,
...
-      // When unspecified, default to fp32 matmul.
-      matmulIOType = DataType::Float;
-
-    ErrorObject status =
-        benchmarkMatmul(matmulM, matmulN, matmulK, transA, transB, iter, dump,
-                        matmulIOType, deviceId);
-
+    ErrorObject status = runMatmulBenchmark(matmulOpts, iter, deviceId, dump);
     if (isError(status)) {
       std::cerr << "Fusilli Benchmark failed: " << status << std::endl;
```
sjain-stanford:
```suggestion
      std::cerr << "Fusilli Matmul Benchmark failed: " << status << std::endl;
```

benchmarks/driver.cpp
```diff
@@ -457,243 +492,271 @@ benchmarkConvDGrad(int64_t n, int64_t c, int64_t d, int64_t h, int64_t w,
...
+  CLI::App *matmulApp = registerMatmulOptions(mainApp, matmulOpts);
+
+  CLI11_PARSE(mainApp, argc, argv);
+
+  std::cout << "Fusilli Benchmark started..." << std::endl;
+
+  if (convApp->parsed()) {
+    ErrorObject status = runConvBenchmark(convOpts, iter, deviceId, dump);
     if (isError(status)) {
       std::cerr << "Fusilli Benchmark failed: " << status << std::endl;
```
sjain-stanford:
```suggestion
      std::cerr << "Fusilli Conv Benchmark failed: " << status << std::endl;
```

PR #65, opened 2025-12-22
[HipdnnPlugin] Add pointwise operations to plugin

AaronStGeorge:
This PR adds pointwise op capability to the plugin. For example the plugin will now offer the fusilli engine as an option when hipDNN is presented with a graph `conv` -> `add` -> `relu`.


sjain-stanford (approved):
Nice!

sjain-stanford:
The CI did pass earlier but failed on re-run - looks like an artifact of tracking `develop` of rocm-libraries - something changed upstream.

PR #71, opened 2025-12-25
[Fusilli] Added ASM emitter for Layernorm node

a-sidorova:
The third part of https://github.com/iree-org/fusilli/issues/45

### DIFF
--- a/samples/layernorm/layernorm_infer_nchw.cpp
+++ b/samples/layernorm/layernorm_infer_nchw.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 Advanced Micro Devices, Inc.
+// Copyright 2026 Advanced Micro Devices, Inc.
 //
 // Licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
--- a/samples/layernorm/layernorm_infer_nchw_scale_bias.cpp
+++ b/samples/layernorm/layernorm_infer_nchw_scale_bias.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 Advanced Micro Devices, Inc.
+// Copyright 2026 Advanced Micro Devices, Inc.
 //
 // Licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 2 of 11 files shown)


include/fusilli/backend/runtime.h
```diff
@@ -242,6 +242,15 @@ inline ErrorObject Graph::execute(
...
   for (const auto &input : fullGraphInputsSorted_) {
+    // Scalar constants are inlined in the IR and not passed as function
+    // arguments, so they should not be in the variantPack.
+    if (input->isScalar()) {
+      FUSILLI_RETURN_ERROR_IF(variantPack.contains(input),
+                              ErrorCode::VariantPackError,
+                              "Scalar constant tensor found in variantPack");
+      continue;
+    }
+
```
IanWood1: Is this true for all scalars? This seems to be the case for epsilon based on [Cudnn's layer norm sample](https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/samples/cpp/norm/layernorm.cpp).


For example, if you have a pointwise multiply `X * s` where `s` is a scalar value. This would have to be passed in through the variant pack.

sjain-stanford: > Can we make the assumption that our isScalar == is_pass_by_value from cudnn?

Yes, when I added this I checked and `is_pass_by_value` was being used for scalars. So please assume this matches-up with cudnn's is_pass_by_value.

sjain-stanford: cudnn handles scalars in two ways: 
```
    // There are two usecases of pass by value tensors:
    // 1. Fused scalar constants
    // 2. Scalar passed during execution
    // In approach 1, users provide a value to embed into the graph.
    // In approach 2, users set is_pass_by_value boolean and then pass a pointer to scalar value with execute() API.
    // A closed set of types that are allowed to be passed by value.
```

I believe epsilon for layernorm falls in category 1 above (it's a compile time scalar constant) and doesn't need `is_pass_by_value` (or `isScalar` in our case) to be set. This is not part of the variant pack either. 

For the pointwise multiply example Ian shared above, we want it to be category 2 (where `isScalar` is set and it's part of the variant pack). 

Does this make sense @a-sidorova , @IanWood1 ?

a-sidorova:
> I think this needs a rebase on main.

@sjain-stanford I've rebased the branch and resolved merge conflicts. Could you please start review when you have time?

Thank you in advance!

a-sidorova:
@sjain-stanford thanks a lot for the updated year!

PR #83, opened 2026-01-07
[hipdnnPlugin] Add support for matmul

AaronStGeorge:
HipDNN and Fusilli both support matmul operators, this PR adds the necessary wiring to the plugin to translate HipDNN graph to Fusilli graph.


plugins/hipdnn-plugin/test/integration/CMakeLists.txt
```diff
@@ -71,3 +71,37 @@ add_fusilli_plugin_test(
     FUSILLI_PLUGIN_NAME="${FUSILLI_PLUGIN_NAME}"
     FUSILLI_PLUGIN_ENGINE_ID="${FUSILLI_PLUGIN_ENGINE_ID}"
 )
+
```
AaronStGeorge: Irrespective of the upcoming larger level integration testing suite I think it's nice to have a few of these simple "sample" tests for local iteration.

plugins/hipdnn-plugin/src/fusilli_plugin.cpp
```diff
@@ -221,41 +221,48 @@ hipdnnPluginStatus_t hipdnnEnginePluginGetApplicableEngineIds(
     return HIPDNN_PLUGIN_STATUS_SUCCESS;
   }
 
-  // Check for conv_fprop graph (optionally with pointwise ops)
+  // Check for supported graph structure
```
BrianHarrisonAMD: I am not familiar with what restrictions fusilli might have for the operations, but it looks like right now it's checking:

- First node conv or matmul
- If conv, then pre & post padding must match

There is the checks that happen in the frontend to ensure it's a valid operation as well, but that means any valid conv, GEMM, or fused conv / GEMM would be accepted currently by fusilli. Is that correct?
There is quite a few pointwise operations that could be set in a graph (reference [here](https://github.com/ROCm/rocm-libraries/blob/develop/projects/hipdnn/frontend/include/hipdnn_frontend/Types.hpp#L28-L78)), and I would think only a subset are valid for fusilli. 

If there is only a subset of patterns and pointwise support in Fusilli, then we will need to add those checks here.

plugins/hipdnn-plugin/test/integration/matmul/batched_matmul_parameterized.cpp
```diff
@@ -0,0 +1,255 @@
...
+  CpuFpReferenceValidation<float> validator(1e-6f, 1e-6f);
+  EXPECT_TRUE(validator.allClose(expectedOutput, outTensor));
+
+  // Clean up.
+  ASSERT_EQ(hipStreamDestroy(stream), HIPDNN_STATUS_SUCCESS);
+  ASSERT_EQ(hipdnnDestroy(handle), HIPDNN_STATUS_SUCCESS);
+}
+
+INSTANTIATE_TEST_SUITE_P(BatchedMatmul, BatchedMatmulParameterizedTest,
+                         ::testing::ValuesIn(getBatchedMatmulTestCases()));
```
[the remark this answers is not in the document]
BrianHarrisonAMD: I think it's best to have both explicit plugin tests for applicability checks, and integration validation checks, but you likely get some amount of coverage relying on the final integration suite only.

For example:
- If Plugin A doesn't support grouped conv, then having a test at the plugin level to ensure it's not applicable for grouped convs is useful.
    - You get a fast signal, and this can be a unit test without any device code (example [here](https://github.com/ROCm/rocm-libraries/blob/develop/dnn-providers/miopen-provider/tests/engines/plans/TestMiopenConvPlanBuilder.cpp)).
- If hipDNN has grouped conv support through plugin B, then the unified test suite should have a test verifying it works, and is supported.  
    - This is slower, requires device code, and breaks at this level will be harder to troubleshoot.
 
In the above scenario, if you don't have the plugin A test and this gets broken, then you will likely get a signal indicating a validation failure or a graph builder error, but it's now further away from the source, and harder to troubleshoot.

Ultimately, each plugin can decide for themselves, but I prefer the fast running minimal checks that don't rely on output validation when & where possible. Having those allows us to have a fast signal for pre-submit, and prevents a lot of churn from issues being found in the slower suites later on.

sjain-stanford (approved):
LGTM, please land once @BrianHarrisonAMD approves.

BrianHarrisonAMD (approved):
LGTM.

My feedback can be split into future PRs, or handled as you see fit.
I don't think it's blocking for this to be merged.

---

I am not sure how you wanted to track operation support for the fusilli plugin ([here](https://github.com/ROCm/rocm-libraries/blob/develop/projects/hipdnn/docs/OperationSupport.md) & [here](https://github.com/ROCm/rocm-libraries/blob/develop/dnn-providers/miopen-provider/docs/OperationSupport.md) is how we are doing it for hipDNN thus far), but you may also want to update the operation support documentation if / when it's available to include the new operation support.

At some point for sure we will want to have an up to date doc that outlines support for each engine, and we will be adding that to the release documentation.

PR #85, opened 2026-01-08
Add reduction attr and node

IanWood1:
This PR implements a `ReductionAttr` class and `ReductionNode` class. The verifier and inference for `ReductionNode` is implemented, but the ASM emitter is left for a follow-up PR.

### DIFF
--- a/include/fusilli/attributes/reduction_attributes.h
+++ b/include/fusilli/attributes/reduction_attributes.h
@@ -25,8 +25,17 @@
 namespace fusilli {
 
 #define FUSILLI_REDUCTION_MODES(OP)                                            \
-  OP(NOT_SET), OP(SUM), /* OP(ADD), OP(MUL), */ OP(MIN),                       \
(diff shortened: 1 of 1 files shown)


include/fusilli/node/reduction_node.h
```diff
@@ -0,0 +1,105 @@
...
+                            "Reduction operation requires X input");
+
+    // Validate output Y exists
+    FUSILLI_RETURN_ERROR_IF(!reductionAttr.getY(), ErrorCode::AttributeNotSet,
+                            "Reduction operation requires Y output");
+
+    // Validate that input and output have the same rank
+    const auto &xTensor = reductionAttr.getX();
+    const auto &yTensor = reductionAttr.getY();
+    if (!xTensor->getDim().empty() && !yTensor->getDim().empty()) {
```
sjain-stanford: xTensor's dim/stride shouldn't be optional. We need it specified to infer the yTensor's shape/stride.

PR #86, opened 2026-01-08
Fix mixed precision matmul failure

IanWood1:
Mixed precision matmuls fail to compile because `torch.aten.matmul` does not support mixed precision operands. This change introduces a cast when the dtypes differ from the compute dtype. This also requires changing the `torch.aten.matmul` asm so that the inputs/results all match the compute data type. The result of the matmul is then cast back to the output tensor's data type.

### DIFF
--- a/include/fusilli/node/matmul_node.h
+++ b/include/fusilli/node/matmul_node.h
@@ -83,7 +83,7 @@ public:
   std::string getResultTypesAsm() const;
   std::string getPermuteAndCastAOpsAsm() const;
   std::string getPermuteAndCastBOpsAsm() const;
-  std::string getPermuteCOpsAsm() const;
+  std::string getPermuteAndCastCOpsAsm() const;
 
   const std::string &getName() const override final {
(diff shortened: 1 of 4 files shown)


sjain-stanford (changes requested):
Just at the outset, I'm wary about introducing dtype casting behavior outside of the compiler (in Fusilli). Fusilli shouldn't be at the liberty to make type casting decisions since these are usually semantic altering, and rather suited for the compiler. I'd look into why there's a restriction on the torch matmul op, and if that can't be lifted, maybe its there for a good reason (and we mimic the same in Fusilli). IMO Fusilli ops should be 1:1 with torch ops with all metadata (shapes, types) explicit and aligned. 

Also, right now all the mixed precision matmuls in prod shapes are from batch matmul (which does support mixed precision) so we probably don't need this change right now. If this becomes a problem, we'd need to expose explicit cast ops through Fusilli API and have the user model it in the created graph.

PR #93, opened 2026-01-12
[FusilliPlugin] Check for unsupported pointwise nodes in graph

AaronStGeorge:
This PR ensures that the plugin doesn't provide an engine if it can't convert pointwise operations in the graph in question. The plugin API `hipdnnEnginePluginGetApplicableEngineIds` is responsible for offering the fusilli plugin engine as a candidate for compile+execution of a given graph. Previously the plugin didn't check for unsupported pointwise mode types and just assumed that any pointwise node was OK.

### DIFF
--- a/plugins/hipdnn-plugin/test/test_fusilli_plugin_api.cpp
+++ b/plugins/hipdnn-plugin/test/test_fusilli_plugin_api.cpp
@@ -10,6 +10,11 @@
 #include <hipdnn_data_sdk/data_objects/data_types_generated.h>
 #include <hipdnn_data_sdk/data_objects/engine_config_generated.h>
 #include <hipdnn_data_sdk/data_objects/pointwise_attributes_generated.h>
+#include <hipdnn_frontend/Graph.hpp>
+#include <hipdnn_frontend/Utilities.hpp>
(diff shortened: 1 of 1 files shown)


plugins/hipdnn-plugin/src/fusilli_plugin.cpp
```diff
@@ -262,6 +262,30 @@ hipdnnPluginStatus_t hipdnnEnginePluginGetApplicableEngineIds(
...
+    const hipdnn_data_sdk::data_objects::PointwiseAttributes *pointwiseAttrs =
+        node.attributes_as_PointwiseAttributes();
+    hipdnn_data_sdk::data_objects::PointwiseMode mode =
+        pointwiseAttrs->operation();
+    if (fusilli::isError(hipDnnPointwiseModeToFusilliMode(mode))) {
+      HIPDNN_LOG_INFO("Fusilli doesn't currently support pointwise Mode {}",
+                      mode);
+      return HIPDNN_PLUGIN_STATUS_SUCCESS;
+    }
+  }
```
BrianHarrisonAMD: Nice that looks good!

PR #98, opened 2026-01-16
Enhance support for pointwise operators

rsuderman:
This adds ll equivalent pointwise operators to fusilli. It also includes implementation and testing for:
- CEIL
- SIGMOID_FWD
- TANH_FWD
- CMP_EQ
- CMP_NEQ
- CMP_LT
- CMP_LE
- CMP_GT
- CMP_GE

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -38,6 +38,7 @@ add_fusilli_tests(
     test_layernorm_attributes.cpp
     test_matmul_attributes.cpp
     test_pointwise_attributes.cpp
+    test_reduction_attributes.cpp
   DEPS
     libfusilli
     Catch2::Catch2WithMain)
@@ -50,6 +51,7 @@ add_fusilli_tests(
     test_layernorm_node.cpp
     test_matmul_node.cpp
     test_pointwise_node.cpp
+    test_reduction_node.cpp
   DEPS
     libfusilli
     Catch2::Catch2WithMain)
--- a/include/fusilli.h
+++ b/include/fusilli.h
@@ -30,6 +30,7 @@
 #include "fusilli/attributes/layernorm_attributes.h" // IWYU pragma: export
 #include "fusilli/attributes/matmul_attributes.h"    // IWYU pragma: export
 #include "fusilli/attributes/pointwise_attributes.h" // IWYU pragma: export
+#include "fusilli/attributes/reduction_attributes.h" // IWYU pragma: export
 #include "fusilli/attributes/tensor_attributes.h"    // IWYU pragma: export
 #include "fusilli/attributes/types.h"                // IWYU pragma: export
 
@@ -39,6 +40,7 @@
 #include "fusilli/node/matmul_node.h"    // IWYU pragma: export
 #include "fusilli/node/node.h"           // IWYU pragma: export
 #include "fusilli/node/pointwise_node.h" // IWYU pragma: export
+#include "fusilli/node/reduction_node.h" // IWYU pragma: export
 
 // Backend:
 #include "fusilli/backend/backend.h" // IWYU pragma: export
--- a/include/fusilli/attributes/reduction_attributes.h
+++ b/include/fusilli/attributes/reduction_attributes.h
@@ -1,0 +1,87 @@
+// Copyright 2026 Advanced Micro Devices, Inc.
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//===----------------------------------------------------------------------===//
+//
+// This file contains attributes (compile-time constant metadata) for
+// reduction nodes.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef FUSILLI_ATTRIBUTES_REDUCTION_ATTRIBUTES_H
+#define FUSILLI_ATTRIBUTES_REDUCTION_ATTRIBUTES_H
+
+#include "fusilli/attributes/attributes.h"
+#include "fusilli/attributes/tensor_attributes.h"
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+namespace fusilli {
+
+#define FUSILLI_REDUCTION_MODES(OP)                                            \
+  OP(NOT_SET)                                                                  \
+  OP(SUM)                                                                      \
+  /* OP(ADD), */                                                               \
+  /* OP(MUL), */                                                               \
+  OP(MIN)                                                                      \
+  OP(MAX)                                                                      \
+  /* OP(AMAX), */                                                              \
+  /* OP(AVG), */                                                               \
+  /* OP(NORM1), */                                                             \
+  /* OP(NORM2), */                                                             \
+  /* OP(MUL_NO_ZEROS) */
+
+class ReductionAttr : public AttributesCRTP<ReductionAttr> {
+public:
+  // Names for Tensor Inputs and Outputs. Reduction has a single input.
+  enum class InputNames : uint8_t { X };
+  enum class OutputNames : uint8_t { Y };
+
+  enum class Mode : uint8_t {
+#define FUSILLI_REDUCTION_MODE_ENUM(mode) mode,
+    FUSILLI_REDUCTION_MODES(FUSILLI_REDUCTION_MODE_ENUM)
+#undef FUSILLI_REDUCTION_MODE_ENUM
+  };
+
+  std::unordered_map<InputNames, std::shared_ptr<TensorAttr>> inputs;
+  std::unordered_map<OutputNames, std::shared_ptr<TensorAttr>> outputs;
+
+  // Setters:
+  FUSILLI_GENERIC_INPUT_TENSOR_SETTER(ReductionAttr, InputNames, X)
+  FUSILLI_GENERIC_OUTPUT_TENSOR_SETTER(ReductionAttr, OutputNames, Y)
+
+  ReductionAttr &setMode(Mode mode) {
+    mode_ = mode;
+    return *this;
+  }
+
+  // Getters:
+  FUSILLI_GENERIC_INPUT_TENSOR_GETTER(InputNames, X)
+  FUSILLI_GENERIC_OUTPUT_TENSOR_GETTER(OutputNames, Y)
+
+  Mode getMode() const { return mode_; }
+
+  // Utilities for reduction modes.
+  static const std::unordered_map<Mode, std::string> kModeToStr;
+
+private:
+  Mode mode_ = Mode::NOT_SET;
+};
+
+inline const std::unordered_map<ReductionAttr::Mode, std::string>
+#define FUSILLI_REDUCTION_MODE_MAP(mode) {ReductionAttr::Mode::mode, #mode},
+    ReductionAttr::kModeToStr = {
+        FUSILLI_REDUCTION_MODES(FUSILLI_REDUCTION_MODE_MAP)};
+#undef FUSILLI_REDUCTION_MODE_MAP
+
+#undef FUSILLI_REDUCTION_MODES
+
+} // namespace fusilli
+
+#endif // FUSILLI_ATTRIBUTES_REDUCTION_ATTRIBUTES_H
--- a/include/fusilli/node/node.h
+++ b/include/fusilli/node/node.h
@@ -36,6 +36,7 @@ public:
     DGrad,
     LayerNorm,
     Matmul,
+    Reduction,
   };
 
   explicit INode(const Context &ctx) : context(ctx) {}
--- a/include/fusilli/node/reduction_node.h
+++ b/include/fusilli/node/reduction_node.h
@@ -1,0 +1,105 @@
+// Copyright 2026 Advanced Micro Devices, Inc.
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//===----------------------------------------------------------------------===//
+//
+// This file contains definitions for the reduction nodes.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef FUSILLI_NODE_REDUCTION_NODE_H
+#define FUSILLI_NODE_REDUCTION_NODE_H
+
+#include "fusilli/attributes/reduction_attributes.h"
+#include "fusilli/attributes/tensor_attributes.h"
+#include "fusilli/graph/context.h"
+#include "fusilli/node/node.h"
+#include "fusilli/support/logging.h"
+
+#include <cstddef>
+#include <cstdint>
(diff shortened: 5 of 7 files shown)


build_tools/cmake/FusilliTestUtils.cmake
```diff
@@ -239,6 +239,45 @@ function(add_fusilli_lit_test)
...
+#    [DEPS <dep> [<dep> ...]]
+#    [TOOLS <tool> [<tool> ...]]
+#  )
+# SRCS
+#  Source files to compile into individual executables (required).
+# DEPS
+#  Library dependencies to be linked to the targets.
+# TOOLS
+#  External tools needed for the tests.
+function(add_fusilli_lit_tests)
```
sjain-stanford: Thanks, this was needed!

include/fusilli/attributes/pointwise_attributes.h
```diff
@@ -71,22 +117,35 @@ class PointwiseAttr : public AttributesCRTP<PointwiseAttr> {
   Mode mode_ = Mode::NOT_SET;
 };
 
+#define DECLARE_STRINGIFY_POINTWISE_MODE(mode)                                 \
+  {PointwiseAttr::Mode::mode, #mode},
+
+// clang-format off
```
sjain-stanford: just curious, what was the issue with clang-format for this bit?

include/fusilli/graph/graph.h
```diff
@@ -690,7 +709,8 @@ Graph::pointwise(const std::shared_ptr<TensorAttr> &in,
   pointwiseAttr.setIN_0(in);
 
   // Set outputs.
-  auto out = outputTensor(pointwiseAttr.getName() + "_OUT_0");
+  DataType outDataType = kPointwiseDefaultDataTypes.at(pointwiseAttr.getMode());
+  auto out = outputTensor(pointwiseAttr.getName() + "_OUT_0", outDataType);
```
sjain-stanford: There's a specific type propagation system in fusilli / cudnn which would be the canonical approach to specify types on outputs - i.e., we don't want to set them in the op builders. This is to be implemented somewhere on the pointwise node's `inferPropertiesNode`. There's a `fillFromContext` call which basically populates unset ones to the default types set on the graph (i.e., IO / intermediate / compute datatypes). But you can also explicitly specify type propagation rules (e.g. use same dtype as inputs or hardcoded to boolean for pointwise comparisons...). In most pointwise samples in cudnn you'll see this explicitly annotated on the outputs by the user, and this is what we're trying to match closely to keep the user interface consistent.

include/fusilli/graph/graph.h
```diff
@@ -458,11 +458,12 @@ class Graph : public INode {
     return ok(true);
   }
 
-  std::shared_ptr<TensorAttr> outputTensor(const std::string &name) {
+  std::shared_ptr<TensorAttr> outputTensor(const std::string &name,
+                                           DataType dt = DataType::NotSet) {
```
sjain-stanford: Same, I'd drop this change (related to the comment about type specification and API consistency with cudnn's [output_tensor](https://sourcegraph.com/github.com/NVIDIA/cudnn-frontend@b372d39879d44c91a8d5b342022e74802b6a8da2/-/blob/include/cudnn_frontend/node_interface.h?L52)).

include/fusilli/attributes/pointwise_attributes.h
```diff
@@ -71,22 +117,35 @@ class PointwiseAttr : public AttributesCRTP<PointwiseAttr> {
...
-        {PointwiseAttr::Mode::NOT_SET, "NOT_SET"},
-        {PointwiseAttr::Mode::RELU_FWD, "RELU_FWD"},
-        {PointwiseAttr::Mode::ADD, "ADD"},
-        {PointwiseAttr::Mode::DIV, "DIV"},
-        {PointwiseAttr::Mode::MUL, "MUL"},
-        {PointwiseAttr::Mode::SUB, "SUB"},
+        FUSILLI_POINTWISE_OPS(DECLARE_STRINGIFY_POINTWISE_MODE)
 };
+// clang-format on
+#undef DECLARE_STRINGIFY_POINTWISE_MODE
```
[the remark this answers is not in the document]
rsuderman: I actually use it in `graph_import.h` to convert the data_sdk values into our fusilli ones. We match naming so it makes things a bit smoother. It also guarantees 1:1 matches.

tests/lit/test_pointwise_asm_emitter_add.cpp
```diff
@@ -11,28 +11,28 @@
 // clang-format off
 //
 // TORCH-CHECK:   module @module {
-// TORCH-CHECK:     func.func @main(%result_: !torch.tensor<[16,256,64,32],f32>, %arg0_input: !torch.vtensor<[16,256,64,32],f32>, %arg1_add: !torch.vtensor<[1,256,1,1],f32>) attributes {torch.assume_strict_symbolic_shapes} {
```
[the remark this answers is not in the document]
rsuderman: I reverted to use the hardcoded names. A few arg names changes due to using the test utilities so I updated to match those. For new tests I will probably stick with regex matching just to avoid line changes when IR is tweaked but only for when exact matching is not checked.

Are there any concerns about name conflicts though? Look a `%var` and  `%var_perm` (or other similar variations) I am a bit worried about name conflict. The names don't seem unique enough such that chaining operations could result in some pretty likely collisions.

include/fusilli/attributes/pointwise_attributes.h
```diff
@@ -24,21 +24,67 @@
 
 namespace fusilli {
 
+#define FUSILLI_POINTWISE_OPS(OP)                                              \
```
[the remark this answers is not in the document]
rsuderman: I commented out any unsupported operations. It should make it easier for adding support rather than adding everywhere else. I mainly added the full list as the copy-paste was easy to add.

tests/CMakeLists.txt
```diff
@@ -94,292 +94,43 @@ add_fusilli_tests(
     Catch2::Catch2WithMain
 )
 
-add_fusilli_lit_test(
-  SRC
+add_fusilli_lit_tests(
```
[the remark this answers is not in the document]
rsuderman: Added some generic comparison tests. They seem pretty general as it is so I just expanded out to having one set specific for comparisons. I will still need to pickup on some of datatype quirks as I *think* I implemented the boolean support correctly but won't know until a higher level integration.

include/fusilli/support/asm_emitter.h
```diff
@@ -1289,106 +1289,87 @@ inline std::string PointwiseNode::getResultNamesAndTypesAsm() const {
...
+{6}
+)";
+
+constexpr std::string_view kSubAddSchema = R"(
+{0}
+{1}
+%alpha_{8} = torch.constant.int 1
+{2}_perm = {7} {3}, %alpha_{8} : {4}, !torch.int -> {5}
+{6}
+)";
```
IanWood1: Maybe place these in `PointwiseNode::emitNodePreAsm`?

include/fusilli/support/asm_emitter.h
```diff
@@ -1289,106 +1289,87 @@ inline std::string PointwiseNode::getResultNamesAndTypesAsm() const {
...
+
 inline std::string PointwiseNode::emitNodePreAsm() const {
   switch (pointwiseAttr.getMode()) {
-  case PointwiseAttr::Mode::RELU_FWD: {
-    constexpr std::string_view schema = R"(
-    {0}
-    {1}_perm = torch.aten.relu {2} : {3} -> {4}
-    {5}
-    )";
+    DECLARE_UNARY_TORCH_EMITTER(CEIL, torch.aten.ceil)
```
IanWood1: Nice, much cleaner than before!

include/fusilli/support/asm_emitter.h
```diff
@@ -1289,106 +1289,87 @@ inline std::string PointwiseNode::getResultNamesAndTypesAsm() const {
...
+{2}_perm = {7} {3}, %alpha_{8} : {4}, !torch.int -> {5}
+{6}
+)";
+
+#define DECLARE_UNARY_TORCH_EMITTER(PWOP, OPIR)                                \
+  DECLARE_UNARY_POINTWISE_EMITTER(PWOP, kUnaryTorchSchema, OPIR)
+#define DECLARE_BINARY_TORCH_EMITTER(PWOP, OPIR)                               \
+  DECLARE_BINARY_POINTWISE_EMITTER(PWOP, kBinaryTorchSchema, OPIR)
+#define DECLARE_SUB_ADD_EMITTER(PWOP, OPIR)                                    \
+  DECLARE_BINARY_POINTWISE_EMITTER(PWOP, kSubAddSchema, OPIR)
```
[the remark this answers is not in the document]
rsuderman: Undefed

include/fusilli/attributes/pointwise_attributes.h
```diff
@@ -71,22 +117,32 @@ class PointwiseAttr : public AttributesCRTP<PointwiseAttr> {
...
-        {PointwiseAttr::Mode::RELU_FWD, "RELU_FWD"},
-        {PointwiseAttr::Mode::ADD, "ADD"},
-        {PointwiseAttr::Mode::DIV, "DIV"},
-        {PointwiseAttr::Mode::MUL, "MUL"},
-        {PointwiseAttr::Mode::SUB, "SUB"},
-};
+        FUSILLI_POINTWISE_OPS(DECLARE_STRINGIFY_POINTWISE_MODE)};
+#undef DECLARE_STRINGIFY_POINTWISE_MODE
+
+// clang-format off
```
sjain-stanford: nit: should we drop the clang-format around this (doesn't seem like it needs it)

include/fusilli/support/asm_emitter.h
```diff
@@ -1289,106 +1289,90 @@ inline std::string PointwiseNode::getResultNamesAndTypesAsm() const {
...
+                       getResultTypesAsm(),              /* {4} */             \
+                       getPermuteOut0OpsAsm(),           /* {5} */             \
+                       #OPIR,                            /* {6} */             \
+                       getName()                         /* {7} */             \
+    );                                                                         \
+  }
+
+#define FUSILLI_DECLARE_BINARY_POINTWISE_EMITTER(PWOP, SCHEMA, OPIR)           \
+  case PointwiseAttr::Mode::PWOP: {                                            \
+    return std::format(SCHEMA, getPermuteInputOpsAsm(0), /* {0} */             \
```
sjain-stanford: same

include/fusilli/support/asm_emitter.h
```diff
@@ -1289,106 +1289,90 @@ inline std::string PointwiseNode::getResultNamesAndTypesAsm() const {
...
+%alpha_{8} = torch.constant.int 1
+{2}_perm = {7} {3}, %alpha_{8} : {4}, !torch.int -> {5}
+{6}
+)";
+
+#define FUSILLI_DECLARE_UNARY_TORCH_EMITTER(PWOP, OPIR)                        \
+  FUSILLI_DECLARE_UNARY_POINTWISE_EMITTER(PWOP, kUnaryTorchSchema, OPIR)
+#define FUSILLI_DECLARE_BINARY_TORCH_EMITTER(PWOP, OPIR)                       \
+  FUSILLI_DECLARE_BINARY_POINTWISE_EMITTER(PWOP, kBinaryTorchSchema, OPIR)
+#define FUSILLI_DECLARE_SUB_ADD_EMITTER(PWOP, OPIR)                            \
```
sjain-stanford:
```suggestion
#define FUSILLI_DECLARE_SUB_ADD_TORCH_EMITTER(PWOP, OPIR)                            \
```

tests/lit/test_pointwise_asm_emitter_add.cpp
```diff
@@ -42,64 +42,20 @@
 //
 // clang-format on
 
+#include "../utils.h"
```
sjain-stanford: If possible consider avoiding a relative import here by taking a direct dep on `libutils` https://sourcegraph.com/github.com/iree-org/fusilli/-/blob/CMakeLists.txt?L206

tests/CMakeLists.txt
```diff
@@ -1,14 +1,17 @@
 # Copyright 2025 Advanced Micro Devices, Inc.
 #
-# Licensed under the Apache License v2.0 with LLVM Exceptions.
-# See https://llvm.org/LICENSE.txt for license information.
-# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Licensed under the Apache License v2.0 with LLVM Exceptions. See
+# https://llvm.org/LICENSE.txt for license information. SPDX-License-Identifier:
+# Apache-2.0 WITH LLVM-exception
```
sjain-stanford: Something happened here. Revert this change?

include/fusilli/graph/graph.h
sjain-stanford: > Do we need to support overriding the output type though? If so I will need to change it to include a conversion operator for the output as torch only supports i1 for comparison operations.

No we shouldn't have to. Let's stick to what torch supports.

tests/CMakeLists.txt
```diff
@@ -26,366 +29,117 @@ if(FUSILLI_SYSTEMS_AMDGPU)
...
-  TOOLS
-    FileCheck
-    iree-opt
-    iree-compile
-)
+add_fusilli_lit_tests(
+  SRCS
+  lit/test_asm_emitter.cpp
+  lit/test_conv_asm_emitter_nchw_kcrs.cpp
+  lit/test_conv_asm_emitter_nchw_krsc.cpp
```
sjain-stanford: same (indent)

sjain-stanford (approved):
LGTM modulo minor comments. Thanks for taking this up.

include/fusilli/support/asm_emitter.h
```diff
@@ -1289,106 +1289,90 @@ inline std::string PointwiseNode::getResultNamesAndTypesAsm() const {
   return getResultNamesAsm() + ": " + getResultTypesAsm();
 }
 
+#define FUSILLI_DECLARE_UNARY_POINTWISE_EMITTER(PWOP, SCHEMA, OPIR)            \
+  case PointwiseAttr::Mode::PWOP: {                                            \
+    return std::format(SCHEMA, getPermuteInputOpsAsm(0), /* {0} */             \
```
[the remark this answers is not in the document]
rsuderman: Unfortunately this violates the clang-format rules. If its a big enough annoyance we can disable clang-format but seemed better to just submit to the format script.

samples/pointwise/pointwise_unary_ops.cpp
```diff
@@ -132,7 +161,7 @@ TEST_CASE("Pointwise unary ops", "[pointwise][graph]") {
 #endif
 
   // int32
-  execute(handlePtr, DataType::Int32, int(-128));
+  // execute(handlePtr, DataType::Int32, int(-128));
```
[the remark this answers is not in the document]
rsuderman: Re-enabled.

tests/CMakeLists.txt
rsuderman: `cmake-format` was going rogue. I disable it for now.

PR #99, opened 2026-01-16
[NFC] Deduplicate ASM transpose emitters

IanWood1:
Removes the `getPermute*OpsAsm` methods for matmul node and all conv nodes. These methods are replaced with a call to the new helper `getPermuteOpsAsm()`.

Pointwise nodes have a slightly different format for the variable names, and to prevent conflicts with https://github.com/iree-org/fusilli/pull/98 they were left unchanged by this commit.

### DIFF
--- a/include/fusilli/node/conv_node.h
+++ b/include/fusilli/node/conv_node.h
@@ -76,9 +76,6 @@ public:
   std::string getStrideOpsAsm() const;
   std::string getPaddingOpsAsm() const;
   std::string getDilationOpsAsm() const;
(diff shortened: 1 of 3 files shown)


sjain-stanford:
> Pointwise nodes have a slightly different format for the variable names, and to prevent conflicts with https://github.com/iree-org/fusilli/pull/98 they were left unchanged by this commit.

Curious, what was the difference here? Can it be consolidated to pointwise nodes as well, now that #98 has landed?

PR #107, opened 2026-01-22
Add reduction ASM emitter

IanWood1:


include/fusilli/support/asm_emitter.h
```diff
@@ -1293,6 +1293,133 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
...
+}
+
+inline std::string ReductionNode::emitNodePreAsm() const {
+  const auto &xT = reductionAttr.getX();
+  const auto &yT = reductionAttr.getY();
+
+  // Compute which dimensions to reduce
+  // Input and output must have the same rank
+  // Find dimensions where Y is 1 and X is greater than 1 (reduced dimensions)
+  std::vector<int64_t> reductionDims;
```
sjain-stanford: Do we want to catch and error on cases where `reductionDims` is empty (say when all dims of X and Y match exactly - indicating a user error). If yes, maybe move this to a common utility that is called from `postValidateNode`. 

Another option is to just assert here but that will only work for debug builds.
```cpp
assert(!reductionDims.empty() && "Reduction requires at least one dimension to reduce");
```

samples/reduction/reduction_ops.cpp
```diff
@@ -0,0 +1,195 @@
...
+      std::format("reduction_{}_dt{}", ReductionAttr::kModeToStr.at(mode),
+                  kDataTypeToMlirTypeAsm.at(type));
+  name += "_x";
+  for (const auto &d : xDim) {
+    name += std::format("_{}", d);
+  }
+  name += "_y";
+  for (const auto &d : yDim) {
+    name += std::format("_{}", d);
+  }
```
sjain-stanford: nit: skip braces
```suggestion
  for (const auto &d : xDim)
    name += std::format("_{}", d);

  name += "_y";
  for (const auto &d : yDim)
    name += std::format("_{}", d);
```

sjain-stanford (approved):
LGTM modulo a few comments. Thanks for adding support for this.

PR #115, opened 2026-01-26
 Add support for custom compiler flags via FUSILLI_EXTRA_COMPILER_FLAGS

bangtianliu:
This PR adds support for passing custom IREE compiler flags to fusilli via:
  - `FUSILLI_EXTRA_COMPILER_FLAGS` environment variable (for C++ driver)
  - `--Xiree-compile` flag (for Python benchmark wrapper)

  This enables users to customize compilation with arbitrary compiler flags,
  such as optimization levels, tuning specs, or debug options.

 Tuning specs provided by tuner can be passed through the flag to enable the performance improvement through tuning. 

Local testing shows **~8.2% performance improvement** on bf16 matrix multiplication (`matmul -M 8192 -N 2048 -K 4096 --transA --a_type bf16 --b_type bf16 --out_type bf16`)  when using tuned transform dialect specs.

**Assisted-by:** Claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,28 @@
 
 [![CI - fusilli](https://github.com/iree-org/fusilli/actions/workflows/build-and-test.yml/badge.svg?branch=main)](https://github.com/iree-org/fusilli/actions/workflows/build-and-test.yml)
 
-Fusilli is a C++ Graph API and JIT Frontend for IREE that leverages just-in-time compiled and code-generated kernels to accelerate training and inference workloads. Inspired by cuDNN's graph API, it exposes cuDNN-like primitives but is backed by the power of the IREE compiler and runtime stack.
+Fusilli is a C++ Graph API and JIT Frontend for IREE that leverages
+just-in-time compiled and code-generated kernels to accelerate training and
+inference workloads. Inspired by cuDNN's graph API, it exposes cuDNN-like
+primitives but is backed by the power of the IREE compiler and runtime stack.
 
-We believe hand-authored GPU kernel libraries are great for highly tuned performance but they are difficult to scale to different models or target architectures and painful to package and release efficiently. This project is founded on the overarching goal to complement the ecosystem of ML frameworks and libraries with a JIT solution, while being competitive to hand-authored kernel libraries. Apart from the core benefit of having a compiler-backed JIT engine that gets progressively and pervasively better, a systemic benefit of this is it helps reduce build times and binary sizes, making it easier to ship software effectively.
+We believe hand-authored GPU kernel libraries are great for highly tuned
+performance but they are difficult to scale to different models or target
+architectures and painful to package and release efficiently. This project is
+founded on the overarching goal to complement the ecosystem of ML frameworks
+and libraries with a JIT solution, while being competitive to hand-authored
+kernel libraries. Apart from the core benefit of having a compiler-backed JIT
+engine that gets progressively and pervasively better, a systemic benefit of
+this is it helps reduce build times and binary sizes, making it easier to ship
+software effectively.
 
 > [!WARNING]
-> :construction: Fusilli is in early stages of development. The operator coverage is limited but growing. APIs may change. :construction:
+> :construction: Fusilli is in early stages of development. The operator
+> coverage is limited but growing. APIs may change. :construction:
 
 > [!NOTE]
-> The name 'Fusilli' is inspired by the term 'fusion' - a bread-and-butter compiler optimization for improving performance.
+> The name 'Fusilli' is inspired by the term 'fusion' - a bread-and-butter
+> compiler optimization for improving performance.
 
 ![Fusilli](docs/fusilli.png)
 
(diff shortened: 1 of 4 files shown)


benchmarks/test_tuning_spec.mlir
```diff
@@ -0,0 +1,7 @@
+// Minimal no-op tuning spec for testing purposes
+// This spec does not modify any operations, just passes them through
```
kuhar:
```suggestion
// Minimal no-op tuning spec for testing purposes.
// This spec does not modify any operations, just passes them through.
```

include/fusilli/backend/backend.h
```diff
@@ -254,6 +276,22 @@ inline std::span<const std::string> getBackendFlags(Backend backend) {
         // clang-format on
     };
 
+    // Helper lambda to add extra compiler flags from environment variable
+    auto addExtraFlags = [](std::vector<std::string> &backendFlags) {
+      if (const char *extraFlags =
+              std::getenv("FUSILLI_EXTRA_COMPILER_FLAGS")) {
+        FUSILLI_LOG_LABEL_ENDL("INFO: Adding extra compiler flags from "
+                               "FUSILLI_EXTRA_COMPILER_FLAGS");
+        auto parsedFlags = parseCompilerFlags(extraFlags);
```
kuhar: spell this out?

benchmarks/run_benchmark.py
```diff
@@ -127,6 +127,7 @@ def run_profiled_command(
     verbose: bool,
     cmd_num: int,
     timeout: int,
+    extra_iree_flags: list[str] | None = None,
```
sjain-stanford: naming nit: for consistency with the env variable `FUSILLI_EXTRA_COMPILER_FLAGS`
```suggestion
    extra_compiler_flags: list[str] | None = None,
```

include/fusilli/backend/backend.h
```diff
@@ -254,6 +278,22 @@ inline std::span<const std::string> getBackendFlags(Backend backend) {
...
+                               "FUSILLI_EXTRA_COMPILER_FLAGS");
+        std::vector<std::string> parsedFlags = parseCompilerFlags(extraFlags);
+        backendFlags.insert(backendFlags.end(), parsedFlags.begin(),
+                            parsedFlags.end());
+      }
+    };
+
+    // Add extra flags to both CPU and AMDGPU backends.
+    addExtraFlags(cpuFlags);
+    addExtraFlags(amdGpuFlags);
```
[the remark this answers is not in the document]
kuhar: +1, I don't understand the distinction -- I don't think the compiler c api has separate entry points for these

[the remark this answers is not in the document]
sjain-stanford: > I don't think the compiler c api has separate entry points for these

OK then it's probably fine to say this is user responsibility (to make sure the flags provided match the backend selected). 

> I don't understand the distinction

The reason we have a map that stores the cpuFlags and amdGpuFlags separately is to allow multiple sessions+graphs to co-exist. So say there's a machine with AMDGPU, users might create graph 1 on the cpu handle and graph 2 on the gpu handle. In such cases, they'd want to make sure the extra flags are not commingled for both backends.

README.md
```diff
@@ -90,6 +90,113 @@ To save the benchmark results as csv, specify `--output-format csv` instead.
...
+- `--verbose`: Enable verbose output
+- `-t, --timeout`: Timeout in seconds per command (default: 30)
+
+#### Custom Compiler Flags
+
+You can pass custom IREE compiler flags using the `FUSILLI_EXTRA_COMPILER_FLAGS` environment variable or the `--Xiree-compile` flag with the Python benchmark wrapper.
+
+**Using the C++ benchmark driver with environment variable:**
+```shell
+FUSILLI_COMPILE_BACKEND_USE_CLI=1 \
```
sjain-stanford: Can drop this line (it is already there in the tuning specs example)

README.md
```diff
@@ -90,6 +90,113 @@ To save the benchmark results as csv, specify `--output-format csv` instead.
...
+FUSILLI_EXTRA_COMPILER_FLAGS="--iree-opt-level=O3" \
+  build/bin/benchmarks/fusilli_benchmark_driver --iter 100 \
+  matmul -M 8192 -N 2048 -K 4096 --transA \
+  --a_type bf16 --b_type bf16 --out_type bf16
+```
+
+**Using the Python benchmark wrapper:**
+```shell
+FUSILLI_COMPILE_BACKEND_USE_CLI=1 \
```
sjain-stanford: Can drop this line (it is already there in the tuning specs example)

README.md
```diff
@@ -90,6 +90,113 @@ To save the benchmark results as csv, specify `--output-format csv` instead.
...
+  -f commands.txt
+```
+
+**Combining tuning specs with other compiler flags:**
+```shell
+python benchmarks/run_benchmark.py \
+  --Xiree-compile="--iree-codegen-tuning-spec-path=/path/to/spec.mlir" \
+  --Xiree-compile="--iree-opt-level=O3" \
+  -f commands.txt -o results.csv
+```
```
sjain-stanford: Let's drop this part since it's a repetition of `Passing multiple compiler flags:` above.

tests/test_backend.cpp
```diff
@@ -7,11 +7,112 @@
...
+  }
+
+  SECTION("Empty string") {
+    std::vector<std::string> result = parseCompilerFlags("");
+    REQUIRE_THAT(result, Catch::Matchers::Equals(std::vector<std::string>{}));
+  }
+
+  SECTION("Whitespace-only string") {
+    std::vector<std::string> result = parseCompilerFlags("   \t  \n  ");
+    REQUIRE_THAT(result, Catch::Matchers::Equals(std::vector<std::string>{}));
```
kuhar: also here

tests/test_backend.cpp
```diff
@@ -7,11 +7,112 @@
...
+  }
+
+  SECTION("Single flag") {
+    std::vector<std::string> result = parseCompilerFlags("--iree-opt-level=O3");
+    REQUIRE_THAT(result, Catch::Matchers::Equals(
+                             std::vector<std::string>{"--iree-opt-level=O3"}));
```
kuhar: Could we make this less verbose by comparing to an initializer list? There's also the `RangeEquals` matcher if `Equals` doesn't work.

```suggestion
    REQUIRE_THAT(result, RangeEquals({"--iree-opt-level=O3"});
```

include/fusilli/backend/backend.h
```diff
@@ -223,6 +224,29 @@ inline std::string getIreeHipTargetForAmdgpu() {
   return arch;
 }
 
+// Parses space-separated compiler flags from a string.
+// Supports double-quote quoting for flags with spaces using std::quoted.
+// Examples:
+//   "--flag1 --flag2=value"
+//   "--flag1 \"--flag2=value with spaces\""
```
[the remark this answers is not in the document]
bangtianliu: No, single quotes do not work as delimiters. The input:
`  "--flag1 '--flag2=value with spaces'"
`
  Would fail and parse incorrectly as:
`  {"--flag1", "'--flag2=value", "with", "spaces'"}
`
  Reason: std::quoted only recognizes double quotes (") as delimiters by default. Single quotes (') are treated as literal characters, not quote delimiters, so the spaces inside them still act as token separators.

  Correct usage for values with spaces:
`  "--flag1 \"--flag2=value with spaces\""
`
  This parses correctly as:
`  {"--flag1", "--flag2=value with spaces"}
`
  I've updated the comment in backend.h to clarify this limitation and added a test case demonstrating the correct usage.

bangtianliu:
cc @kuhar for review

tests/test_backend.cpp
```diff
@@ -12,6 +12,98 @@
 
 using namespace fusilli;
 
+//===----------------------------------------------------------------------===//
+// Tests for parseCompilerFlags
+//===----------------------------------------------------------------------===//
+
+TEST_CASE("parseCompilerFlags", "[backend][flags]") {
```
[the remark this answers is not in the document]
kuhar: Ah, so they do know how to pretty print these without marchers. thanks

PR #122, opened 2026-01-29
[CMake] Fix clang-tidy integration for non-Docker builds

IanWood1:
When `-DFUSILLI_ENABLE_CLANG_TIDY=ON` is used outside the docker setup, builds failed with clang-tidy errors from Catch2.  This occurred because `CMAKE_CXX_CLANG_TIDY` was set globally, causing clang-tidy to analyze all targets created after it was set, including third-party libraries added via `FetchContent`. CI builds worked because Catch2 was pre-installed in the Docker image, so it wasn't built from source and clang-tidy never ran on it.

Changes:
- Instead of using global `CMAKE_CXX_CLANG_TIDY`, clang-tidy is now enabled explicitly on project-owned targets only using the `CXX_CLANG_TIDY` target property
- Fixed fusilli_find_program(): Added check to prevent creating duplicate imported targets when called multiple times
- Disabled `llvm-header-guard` check: This check was causing errors in sample files and the check is only really intended to work with LLVM. See https://github.com/nod-ai/amd-shark-ai/pull/2607#discussion_r2470416426 for details.
- Updated `samples/layernorm/layernorm_utils.h` and `tests/hip_tests/hip_kernels.h` to use consistent `FUSILLI_` prefix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,28 @@
 
 [![CI - fusilli](https://github.com/iree-org/fusilli/actions/workflows/build-and-test.yml/badge.svg?branch=main)](https://github.com/iree-org/fusilli/actions/workflows/build-and-test.yml)
 
-Fusilli is a C++ Graph API and JIT Frontend for IREE that leverages just-in-time compiled and code-generated kernels to accelerate training and inference workloads. Inspired by cuDNN's graph API, it exposes cuDNN-like primitives but is backed by the power of the IREE compiler and runtime stack.
+Fusilli is a C++ Graph API and JIT Frontend for IREE that leverages
+just-in-time compiled and code-generated kernels to accelerate training and
+inference workloads. Inspired by cuDNN's graph API, it exposes cuDNN-like
+primitives but is backed by the power of the IREE compiler and runtime stack.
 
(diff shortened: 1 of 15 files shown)


build_tools/cmake/FusilliBuildUtils.cmake
```diff
@@ -5,6 +5,33 @@
...
+    return()
+  endif()
+
+  # Ensure clang-tidy is available
+  fusilli_find_program(clang-tidy INSTALL_INSTRUCTIONS
+    "Please install clang-tidy (e.g., apt install clang-tidy).")
+
+  set_target_properties(${target} PROPERTIES
+    CXX_CLANG_TIDY
+      "clang-tidy;-warnings-as-errors=*;--config-file=${PROJECT_SOURCE_DIR}/.clang-tidy"
```
sjain-stanford: nit: split one per line

build_tools/cmake/FusilliTestUtils.cmake
```diff
@@ -374,4 +374,7 @@ function(_add_fusilli_executable_for_test)
       ${_RULE_NAME} PROPERTIES
       RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/${_RULE_BIN_SUBDIR}
   )
+
+  # Enable clang-tidy.
+  fusilli_enable_clang_tidy(${_RULE_NAME})
```
sjain-stanford: This gets applied to `add_fusilli_samples`,  `add_fusilli_tests` and `add_fusilli_lit_test`. What about `add_fusilli_benchmark`?

IanWood1: `add_fusilli_benchmark` adds tests by running `fusilli_benchmark_driver` with certain flags. `fusilli_enable_clang_tidy(fusilli_benchmark_driver)` should cover checking the c++ benchmark code in benchmarks/CMakeLists.txt.

PR #123, opened 2026-01-29
[Support] Add portable float16 and bfloat16 types for Windows

rsuderman:
Add `float16` and `bfloat16` type support for `msvcc`. Due to `msvcc` not supporting these float bitwidths we include an implementation that relies on the `iree` type conversion operations. Normal operations and implicitly support via implicit conversion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,6 +139,9 @@ set_target_properties(libfusilli PROPERTIES EXPORT_NAME fusilli)
 # FetchContent) and against an installed version (via find_package).
 add_library(fusilli::fusilli ALIAS libfusilli)
 
+# Enable clang-tidy.
+fusilli_enable_clang_tidy(libfusilli)
+
 # INTERFACE libraries don't install/export header files added via
 # `target_include_directories`. Using a file set property ensures headers are
 # available in the installed/exported package.
@@ -165,16 +168,6 @@ if(FUSILLI_CODE_COVERAGE)
   set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Debug build" FORCE)
 endif()
 
-# Enable after test/samples to avoid catch2 errors.
-if(FUSILLI_ENABLE_CLANG_TIDY)
-  fusilli_find_program(clang-tidy INSTALL_INSTRUCTIONS "Please install clang-tidy (e.g., apt install clang-tidy).")
-  set(CMAKE_CXX_CLANG_TIDY
-    clang-tidy;
-    -warnings-as-errors=*;
-    --config-file=${CMAKE_CURRENT_SOURCE_DIR}/.clang-tidy;
-  )
-endif()
-
 # Enable AMD GPU build if requested
 if(FUSILLI_SYSTEMS_AMDGPU)
   # Check for Linux AMD GPU device node - `/dev/kfd`
@@ -198,6 +191,9 @@ if(FUSILLI_BUILD_TESTS OR FUSILLI_BUILD_BENCHMARKS)
       QUIET
       GIT_REPOSITORY https://github.com/catchorg/Catch2.git
       GIT_TAG        v3.11.0
+      # SYSTEM marks Catch2 headers as system includes, suppressing compiler
+      # warnings and clang-tidy diagnostics from third-party code.
+      SYSTEM
     )
     FetchContent_MakeAvailable(Catch2)
   endif()
@@ -209,6 +205,8 @@ if(FUSILLI_BUILD_TESTS OR FUSILLI_BUILD_BENCHMARKS)
     libfusilli
     Catch2::Catch2WithMain
   )
+  # Enable clang-tidy.
+  fusilli_enable_clang_tidy(libutils)
 endif()
 
 # Build tests and samples
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -16,6 +16,8 @@ set_target_properties(
   fusilli_benchmark_driver PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/benchmarks
 )
+# Enable clang-tidy.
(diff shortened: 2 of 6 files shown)


include/fusilli/support/float_types.h
```diff
@@ -0,0 +1,273 @@
+// Copyright 2025 Advanced Micro Devices, Inc.
```
sjain-stanford:
```suggestion
// Copyright 2026 Advanced Micro Devices, Inc.
```

include/fusilli/support/float_types.h
```diff
@@ -0,0 +1,273 @@
...
+struct Float16 {
+  int16_t data;
+
+  constexpr Float16() : data(0) {}
+
+  // Construct from float (handles double via implicit conversion)
+  constexpr Float16(float f) : data(floatToFp16Bits(f)) {}
+
+  // Convert to float
+  constexpr float toFloat() const { return fp16BitsToFloat(data); }
```
sjain-stanford: `toFloat()`, `toBits()`, and `fromBits()` could benefit from `[[nodiscard]]` to catch accidental misuse:

```cpp
[[nodiscard]] constexpr float toFloat() const { return fp16BitsToFloat(data); }
[[nodiscard]] constexpr int16_t toBits() const { return data; }
[[nodiscard]] static constexpr Float16 fromBits(int16_t bits) { ... }
```

sjain-stanford (commented):
I didn't closely follow the from-scratch fp16 / bf16 implementations but do have a high level comment on considering reusing an existing one (from PyTorch or so) to avoid potential bugs / pitfalls down the road. 

I'd also like to invite @bjacob for a review if possible.

include/fusilli/support/float_types.h
```diff
@@ -0,0 +1,273 @@
...
+    Float16 result;
+    result.data = bits;
+    return result;
+  }
+
+  // Get raw bits
+  constexpr int16_t toBits() const { return data; }
+
+private:
+  static constexpr int16_t floatToFp16Bits(float f) {
```
bjacob: We have conversions between f32 and all narrower float types in IREE runtime here:
https://github.com/iree-org/iree/blob/main/runtime/src/iree/base/internal/math.h#L267-L615

For `f16` that would be: `iree_math_f32_to_f16` and `iree_math_f16_to_f32`.

tests/test_float_types.cpp
```diff
@@ -0,0 +1,444 @@
+// Copyright 2025 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
rsuderman: Updated.

include/fusilli/support/float_types.h
```diff
@@ -0,0 +1,273 @@
...
+
+namespace fusilli {
+
+// IEEE 754 half-precision floating point (Float16)
+// Format: 1 sign bit, 5 exponent bits, 10 mantissa bits
+//
+// This type provides implicit conversions to/from float, allowing seamless
+// interoperability with float arithmetic. All operations are performed in
+// float precision through these conversions.
+struct Float16 {
```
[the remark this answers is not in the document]
rsuderman: I stripped them out and swapped over to the IREE implementation. Initially I thought we required the operators for the samples tests (specifically generating reference implementations) but it turns out we manually upcast to double anyway.

include/fusilli/support/float_types.h
rsuderman: Swapped over to the iree implementations.

include/fusilli/support/float_types.h
```diff
@@ -0,0 +1,114 @@
...
+  uint16_t data;
+
+  constexpr Float16() : data(0) {}
+
+  // Construct from float (handles double via implicit conversion)
+  constexpr Float16(float f) : data(iree_math_f32_to_f16(f)) {}
```
bjacob: Constructors taking 1 parameter should be `explicit` unless you really want the implicit-conversion behaviour.

bjacob (approved):

include/fusilli/support/float_types.h
```diff
@@ -0,0 +1,114 @@
...
+#define FUSILLI_SUPPORT_FLOAT_TYPES_H
+
+#include "fusilli/support/target_platform.h"
+
+#include <iree/base/internal/math.h>
+
+#include <bit>
+#include <cmath>
+#include <cstdint>
+#include <limits>
```
sjain-stanford: Re-evaluate if these are needed after the refactoring to use iree type conversions utilities. Potentially, `bit`, `cmath` and `limits` may be removed?

tests/test_float_types.cpp
```diff
@@ -0,0 +1,294 @@
+// Copyright 2026 Advanced Micro Devices, Inc.
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <fusilli/support/float_types.h>
```
sjain-stanford: nit: Include `<fusilli.h>` in unit tests.

sjain-stanford (approved):
LGTM, thanks for switching over to IREE type conversions. Feel free to land after addressing any remaining comments.

tests/test_float_types.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+#include <catch2/catch_test_macros.hpp>
+#include <cmath>
+#include <limits>
+
+using namespace fusilli;
+
+// =============================================================================
+// Float16 Tests
```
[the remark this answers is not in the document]
bjacob: Yes. maxDenorm is 2^-14 * (1023/1024), as the comment here actually points out.  When cast to float, it becomes normal.  Float has larger exponent range than float16, so to get a subnormal float, you would have to go much farther out than any nonzero float16 (even subnormal) value.

Now that you're using IREE conversions, do you need to still need these tests anyway? They seem redundant with https://github.com/iree-org/iree/blob/main/runtime/src/iree/base/internal/math_test.cc

PR #124, opened 2026-01-30
[Fusilli] Supported build with ASAN

a-sidorova:
Implemented build support with [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) for Unix systems with clang and added their builds in CI.

ASAN support will land in phases. This PR just places the initial boilerplate to iterate on ASAN flags.

### DIFF
--- a/.github/workflows/build-and-test-linux.yml
+++ b/.github/workflows/build-and-test-linux.yml
@@ -99,6 +99,8 @@ jobs:
                  tests/test_cache_empty.sh"
 
     - name: Test fusilli (iree-compile CLI)
+      # Excluded for ASAN runs: the CLI backend spawns `iree-compile`
+      # (not ASAN-instrumented) as a separate process.
       if: ${{ contains(matrix.name, 'debug') || contains(matrix.name, 'release') }}
       run: |
         build_tools/docker/exec_docker_ci.sh \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,11 @@ if(FUSILLI_ENABLE_ASAN)
   list(APPEND FUSILLI_SANITIZER_COMPILE_FLAGS -fsanitize=address -fno-omit-frame-pointer -g -mllvm -asan-globals=0)
   list(APPEND FUSILLI_SANITIZER_LINK_FLAGS -fsanitize=address)
 
+  # Enable recovery mode so the program can continue past ASAN errors,
+  # allowing multiple issues to be reported in a single run. This is
+  # experimental and only supported by some compilers. It takes effect
+  # when `ASAN_OPTIONS=halt_on_error=0` is set at runtime; the default
+  # (halt_on_error=1) still aborts on the first error.
   check_cxx_compiler_flag("-fsanitize-recover=address" SANITIZE_RECOVER_ADDRESS_SUPPORTED)
   if (SANITIZE_RECOVER_ADDRESS_SUPPORTED)
     list(APPEND FUSILLI_SANITIZER_COMPILE_FLAGS -fsanitize-recover=address)
--- a/build_tools/sanitizers/lsan_suppressions.txt
+++ b/build_tools/sanitizers/lsan_suppressions.txt
@@ -1,6 +1,6 @@
 # LSAN suppressions for leaks in external libraries (IREE, ROCm)
 
-# TODO(iree-org/iree#23526): Unsuppress when the IREE memory leak is fixed.
+# TODO(iree-org/iree#23526): Memory leak in IREE HIP driver.
 leak:iree_hal_hip_dispatch_completed_create
 
 # Suppressions from IREE project


include/fusilli/attributes/common.h
```diff
@@ -27,12 +27,15 @@ enum class NormFwdPhase : uint8_t {
...
-};
+inline const std::unordered_map<NormFwdPhase, std::string> &
+normFwdPhaseToStr() {
+  static const std::unordered_map<NormFwdPhase, std::string> map = {
+      {NormFwdPhase::NOT_SET, "NOT_SET"},
+      {NormFwdPhase::TRAINING, "TRAINING"},
+      {NormFwdPhase::INFERENCE, "INFERENCE"},
+  };
+  return map;
+}
```
a-sidorova: Have to move static `std::unordered_map` to accessor function with local static variable because of ODR violation here:
- When this header is included in multiple `.cpp` files (in tests) that get linked into the same binary, each translation unit gets its own copy of these static globals (because of `static`). The string literals (like "f16", "bf16", etc.) inside them produce duplicate `.str` globals, which triggers the ASAN ODR violation.
- We can wrap these maps in inline functions with function-local static variables — these are guaranteed to be initialized exactly once.

CMakeLists.txt
```diff
@@ -67,13 +67,19 @@ option(FUSILLI_CODE_COVERAGE "Enable code coverage for tests" OFF)
...
 message(STATUS "Fusilli supported systems:")
 if(FUSILLI_SYSTEMS_AMDGPU)
   message(STATUS "  - AMD GPU")
 endif()
 message(STATUS "  - Host")
 
+if(FUSILLI_ENABLE_ASAN AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  message(WARNING "AddressSanitizer is only supported with clang. Disabled AddressSanitizer.")
+  set(FUSILLI_ENABLE_ASAN OFF CACHE BOOL "Enable AddressSanitizer" FORCE)
+endif()
```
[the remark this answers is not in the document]
a-sidorova: Let's add error to make it more visible. Thanks

.github/workflows/build-and-test-linux.yml
```diff
@@ -67,6 +67,32 @@ jobs:
               -DFUSILLI_CODE_COVERAGE=ON
               -DFUSILLI_SYSTEMS_AMDGPU=OFF
               -DFUSILLI_ENABLE_CLANG_TIDY=ON
+          - name: gfx942_clang22_asan
+            runs-on: linux-mi325-1gpu-ossci-iree-org
+            cmake-options:
+              -DCMAKE_C_COMPILER=clang-22
+              -DCMAKE_CXX_COMPILER=clang++-22
+              -DCMAKE_BUILD_TYPE=Debug
+              -DFUSILLI_ENABLE_LOGGING=ON
```
sjain-stanford: Just a thought, if it's ASAN errors we're interested in, we could perhaps disable logging (as long as that doesn't hide any ASAN errors).

include/fusilli/support/asm_emitter.h
```diff
@@ -147,7 +156,7 @@ inline std::string getPermuteOpsAsm(const std::shared_ptr<TensorAttr> &tensor,
   std::string toType = tensor->getTensorTypeAsm(
       /*isValueTensor=*/true, /*useLogicalDims=*/isInput);
 
-  constexpr std::string_view schema = R"(
+  static constexpr char schema[] = R"(
```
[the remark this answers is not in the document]
a-sidorova: Sure, added.

.github/workflows/build-and-test-linux.yml
```diff
@@ -67,6 +67,32 @@ jobs:
...
+              -DFUSILLI_SYSTEMS_AMDGPU=ON
+              -DFUSILLI_ENABLE_ASAN=ON
+            asan-env: >-
+              ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer-18
```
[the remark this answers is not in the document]
a-sidorova: Great! Thanks a lot for this feature. Now the versions are aligned!

CMakeLists.txt
```diff
@@ -67,13 +68,18 @@ option(FUSILLI_CODE_COVERAGE "Enable code coverage for tests" OFF)
...
 
 message(STATUS "Fusilli supported systems:")
 if(FUSILLI_SYSTEMS_AMDGPU)
   message(STATUS "  - AMD GPU")
 endif()
 message(STATUS "  - Host")
 
+if(FUSILLI_ENABLE_ASAN AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  message(FATAL_ERROR "AddressSanitizer requires Clang as the C++ compiler, but '${CMAKE_CXX_COMPILER_ID}' was detected.")
+endif()
```
sjain-stanford: Another check worth adding:
```cmake
if(FUSILLI_ENABLE_ASAN AND FUSILLI_CODE_COVERAGE)
  message(FATAL_ERROR "FUSILLI_ENABLE_ASAN and FUSILLI_CODE_COVERAGE cannot be enabled simultaneously.")
endif()
```

.github/workflows/build-and-test-linux.yml
```diff
@@ -67,6 +67,30 @@ jobs:
...
+            cmake-options:
+              -DCMAKE_C_COMPILER=clang-22
+              -DCMAKE_CXX_COMPILER=clang++-22
+              -DCMAKE_BUILD_TYPE=Debug
+              -DFUSILLI_SYSTEMS_AMDGPU=ON
+              -DFUSILLI_ENABLE_ASAN=ON
+            asan-env: >-
+              ASAN_SYMBOLIZER_PATH=/workspace/.cache/docker/venv/bin/llvm-symbolizer-22
+              ASAN_OPTIONS=detect_leaks=1:halt_on_error=0
+              LSAN_OPTIONS=suppressions=/workspace/build_tools/sanitizers/lsan_suppressions.txt
```
sjain-stanford: Instead of setting ASAN env vars in the CI workflow yml, I'd just bake it in the CMake rules (e.g. in `FusilliTestUtils.cmake` and benchmark targets) so that local runs are basically just `ctest ...` without prefixing with env vars. All that the user needs to specify when building is `-DFUSILLI_ENABLE_ASAN=ON`. 

For example `_add_fusilli_ctest_target` can be updated to include:
```cmake
  if(FUSILLI_ENABLE_ASAN)
    list(APPEND _ENV_VARS
      "ASAN_OPTIONS=..."
      "UBSAN_OPTIONS=..."
      "LSAN_OPTIONS=suppressions=${PROJECT_SOURCE_DIR}/build_tools/sanitizers/lsan_suppressions.txt"
    )
  endif()
```

CMakeLists.txt
a-sidorova: I see that they are supported for different compiler. But let's add just in case for future support.

CMakeLists.txt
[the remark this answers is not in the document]
a-sidorova: Let's try and see if there are hidden leaks

.github/workflows/build-and-test-linux.yml
sjain-stanford: We'll need a custom approach for each option since the ergonomics are different. Let's tackle them one at a time:

1. LSAN_OPTIONS / suppressions: This has to be baked in since the suppression file is checked-in to the code. IREE does it this way too ([here](https://sourcegraph.com/github.com/iree-org/iree@604cb5c7a7e77fc4fa62e1531589dca2a4d3dd82/-/blob/build_tools/cmake/iree_native_test.cmake?L76-80)).
2. LLVM_SYMBOLIZER: I am pretty sure ASAN auto-finds `llvm-symbolizer` if on `$PATH` - this should be the case for both CI (with the docker changes to add it to venv) and local env (but users need to make sure it is visible on `$PATH`). At least as of 2013 it should: https://git.raptorcs.com/git/bcm5719-llvm/commit/compiler-rt/lib?id=5b5c99d219884c4356d0ce47b896ed7c4d098cb2
3. ASAN_OPTIONS: You're right that developers might want to tweak these, but a sensible default that the developer can override (by setting the env var in their shell) is strictly better than no default at all. Then CI can use the defaults baked in. Look into CMake's `set_if_undefined` which only takes effect when developer hasn't set them in their shell:

In `_add_fusilli_ctest_target` (`build_tools/cmake/FusilliTestUtils.cmake`), after the existing `set_tests_properties` for `ENVIRONMENT`. The same will be needed in benchmarks but 

```cmake
  # Set environment variables for test
  set_tests_properties(
    ${_RULE_NAME} PROPERTIES
    ENVIRONMENT "${_ENV_VARS}"
  )

  # Set sanitizer environment defaults (overridable by user env vars).
  if(FUSILLI_ENABLE_ASAN)
    set_tests_properties(${_RULE_NAME} PROPERTIES
      ENVIRONMENT_MODIFICATION
        "ASAN_OPTIONS=set_if_undefined:detect_leaks=1:halt_on_error=1"
        "LSAN_OPTIONS=set_if_undefined:suppressions=${PROJECT_SOURCE_DIR}/build_tools/sanitizers/lsan_suppressions.txt"
    )
  endif()
```

sjain-stanford: I created a PR to your branch with a possible approach: https://github.com/a-sidorova/fusilli/pull/1

build_tools/sanitizers/lsan_suppressions.txt
```diff
@@ -0,0 +1,6 @@
+# LSAN suppressions for leaks in external libraries (IREE, ROCm)
+
+leak:iree_hal_hip_dispatch_completed_create
```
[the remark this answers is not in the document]
sjain-stanford: Claude thinks the leak is in IREE and also proposed a fix: https://gist.github.com/sjain-stanford/a3fe37b3ad04665580ecfdf198cf42c9

sjain-stanford (approved):
Most of my comments are resolved / deferred, just a few remain - adding a TODO with the iree issue link on the LSAN suppression, and the `halt_on_error=0` comment. LGTM otherwise.

.github/workflows/build-and-test-linux.yml
```diff
@@ -63,6 +63,25 @@ jobs:
...
+          - name: gfx942_clang22_asan
+            runs-on: linux-mi325-1gpu-ossci-iree-org
+            cmake-options:
+              -DCMAKE_C_COMPILER=clang-22
+              -DCMAKE_CXX_COMPILER=clang++-22
+              -DCMAKE_BUILD_TYPE=Debug
+              -DFUSILLI_SYSTEMS_AMDGPU=ON
+              -DFUSILLI_ENABLE_ASAN=ON
+            asan-env: >-
+              ASAN_OPTIONS=detect_leaks=1:halt_on_error=0
```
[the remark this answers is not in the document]
a-sidorova: My idea was simple: to log all errors that exists there. And after your first comment above I wanted to come back to this question when we fix other things. But thank you so much that you check this moment yourself! Hopefully it didn't take a lot of time 

Let's remove `halt_on_error=0` from CI and README.md.

BTW, Default value is really `1`: https://github.com/google/sanitizers/wiki/AddressSanitizerFlags

PR #133, opened 2026-02-05
FUSILLI_REQUIRE_UNWRAP fixes

sjain-stanford:
https://github.com/iree-org/fusilli/pull/107 had clean CI but was on an older base, and in the meantime FUSILLI_REQUIRE_UNWRAP macro changes from https://github.com/iree-org/fusilli/pull/127 had landed. 

It might be a good idea to require branches to be updated before landing:
image


IanWood1 (approved):
Thank you!

PR #135, opened 2026-02-05
Windows compatibility fixes across the codebase

rsuderman:
- Add MSVC compile options (/bigobj, WIN32_LEAN_AND_MEAN) in CMakeLists.txt
- Use std::map instead of std::unordered_map for deterministic iteration in pointwise attributes
- Add cross-platform cache directory support using SHGetKnownFolderPath on Windows and extract getCacheDir() utility
- Open cache files in binary mode to avoid line-ending corruption
- Use platform-specific IREE compiler library name (.dll vs .so)
- Use platform-specific null redirect (NUL vs /dev/null) and handle \r in python output parsing
- Add FUSILLI_NEWLINE macro for platform-appropriate line endings
- Remove constexpr from Float16/BFloat16 conversion functions for MSVC compat
- Fix f16 literal suffixes to standard f suffix
- Use platform-specific library names in dllib tests (msvcrt.dll vs libm.so.6)
- Rename shadowed 'small' variable to avoid Windows macro conflicts
- Add bounds check in conv output shape inference
- Add break in pointwise stride propagation loop
- Remove printf-based tests not portable to Windows

### DIFF
--- a/include/fusilli/support/python_utils.h
+++ b/include/fusilli/support/python_utils.h
@@ -53,7 +53,8 @@ inline std::vector<std::string> getPythonSitePackages() {
   // Handle both \n and \r for cross-platform compatibility.
   std::string path;
   for (char c : *output) {
-    // Strip out any trailing newlines or carriage returns, and split on newlines.
+    // Strip out any trailing newlines or carriage returns, and split on
+    // newlines.
     if (c == '\n' || c == '\r') {
       if (!path.empty()) {
         sitePaths.push_back(path);


include/fusilli/node/conv_node.h
```diff
@@ -44,6 +44,10 @@ inline std::vector<int64_t> getConvInferredOutputShape(
 
   // N (batch dim)
   yDim[0] = xDim[0];
+
+  if (yDim.size() < 2)
+    return yDim;
```
IanWood1: BTW I created https://github.com/iree-org/fusilli/pull/136 to just remove the test. I don't think we should be calling `inferPropertiesNode` on illegal nodes. @sjain-stanford let me know what you think.

rsuderman: Removed

include/fusilli/support/cache.h
```diff
@@ -101,6 +108,36 @@ class CacheFile {
...
+      cacheDir = std::string(cacheDirC);
+
+#if defined(FUSILLI_PLATFORM_WINDOWS)
+    if (cacheDir.empty()) {
+      PWSTR pathBuf;
+      HRESULT hr =
+          SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, NULL, &pathBuf);
+      if (SUCCEEDED(hr)) {
+        cacheDir = std::filesystem::path(pathBuf).string();
+      }
```
sjain-stanford:
```suggestion
      if (SUCCEEDED(hr))
        cacheDir = std::filesystem::path(pathBuf).string();
```

tests/test_dllib.cpp
```diff
@@ -11,6 +11,14 @@
 #include <cmath>
 #include <utility>
 
+#ifdef FUSILLI_PLATFORM_WINDOWS
+constexpr const char *kLib = "msvcrt.dll";
```
sjain-stanford: naming nit: here and elsewhere
```suggestion
constexpr const char *kLibName = "msvcrt.dll";
```

include/fusilli/attributes/pointwise_attributes.h
```diff
@@ -86,8 +87,8 @@ class PointwiseAttr : public AttributesCRTP<PointwiseAttr> {
   };
 #undef FUSILLI_POINTWISE_DECLARE
 
-  std::unordered_map<InputNames, std::shared_ptr<TensorAttr>> inputs;
-  std::unordered_map<OutputNames, std::shared_ptr<TensorAttr>> outputs;
+  std::map<InputNames, std::shared_ptr<TensorAttr>> inputs;
+  std::map<OutputNames, std::shared_ptr<TensorAttr>> outputs;
```
sjain-stanford: I align with the goal to ensure deterministic iteration order, but changing the container type here is not the best solution because:
- It's applied inconsistently: all op attributes and even the base `AttributeCRTP` class expects derived classes to implement unordered maps for inputs/outputs (https://sourcegraph.com/github.com/iree-org/fusilli/-/blob/include/fusilli/attributes/attributes.h?L30-35 - this closely mimics cudnn's attributes)
- The `InputNames` and `OutputNames` enum have fixed / known set of values that can be iterated explicitly for determinism
- Changes the perf characteristics (O(log n) vs O(1) for lookups)

Instead could we refactor `PointwiseNode::inferPropertiesNode` to iterate over inputs in explicit order using the typed getters:
   ```cpp
   // Current (non-deterministic):
   for (const auto &[_, inTensor] : pointwiseAttr.inputs)
     if (inTensor)
       inputShapes.push_back(inTensor->getDim());

   // Refactored (deterministic, explicit order):
   for (const auto &inTensor : {pointwiseAttr.getIN_0(),
                                pointwiseAttr.getIN_1(),
                                pointwiseAttr.getIN_2()}) {
     if (inTensor)
       inputShapes.push_back(inTensor->getDim());
   }
   ```
(same for stride propagation that follows)

This is also consistent with the ASM emitters which use explicit getters (`getIN_0()`, `getIN_1()`, `getIN_2()`...) for inputs/outputs.

include/fusilli/support/cache.h
sjain-stanford: I think `SHGetKnownFolderPath` allocates memory for `pathBuf` but isn't being freed? Claude tells me the right way to do this is using `CoTaskMemFree(pathBuf)`... 
   ```cpp
   if (cacheDir.empty()) {
     PWSTR pathBuf = nullptr;
     HRESULT hr =
         SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, NULL, &pathBuf);
     if (SUCCEEDED(hr))
       cacheDir = std::filesystem::path(pathBuf).string();
     if (pathBuf)
       CoTaskMemFree(pathBuf);
   }
   ```

include/fusilli/support/cache.h
```diff
@@ -101,6 +108,36 @@ class CacheFile {
...
+    // the GitHub Actions CI runners as well as for LIT tests that rely
+    // on dumping/reading intermediate compilation artifacts to/from disk.
+    const char *cacheDirC = std::getenv("FUSILLI_CACHE_DIR");
+    std::string cacheDir;
+    if (cacheDirC)
+      cacheDir = std::string(cacheDirC);
+
+#if defined(FUSILLI_PLATFORM_WINDOWS)
+    if (cacheDir.empty()) {
+      PWSTR pathBuf;
```
sjain-stanford: Consider `std::wstring` for better support for non-ASCII chars in the user profile path.

include/fusilli/node/conv_node.h
```diff
@@ -44,6 +44,7 @@ inline std::vector<int64_t> getConvInferredOutputShape(
 
   // N (batch dim)
   yDim[0] = xDim[0];
+
```
sjain-stanford: unintentional newline?

tests/test_process.cpp
```diff
@@ -23,13 +23,6 @@ TEST_CASE("execCommand basic functionality", "[support][process]") {
     REQUIRE(result->find("hello") != std::string::npos);
   }
 
-  SECTION("Execute command with multiple lines") {
-    auto result = execCommand("printf 'line1\nline2'");
-    REQUIRE(result.has_value());
-    REQUIRE(result->find("line1") != std::string::npos);
-    REQUIRE(result->find("line2") != std::string::npos);
-  }
-
```
sjain-stanford: why remove?

include/fusilli/support/external_tools.h
```diff
@@ -67,7 +68,11 @@ inline std::string getIreeCompilerLibPath() {
   }
 
   // Fallback: let the system search for it (may be in LD_LIBRARY_PATH).
+#ifdef FUSILLI_PLATFORM_WINDOWS
+  return std::string("IREECompiler.dll");
+#else
   return std::string("libIREECompiler.so");
+#endif
```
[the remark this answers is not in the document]
sjain-stanford: Good point, we can't ensure the order of defines to be sure. In that case let's update all occurrences to error explicitly.

tests/test_process.cpp
rsuderman: Newline behavior for powershell can be a bit weird. Overall it does not matter much for the actual `execCommand` testing so just easier to remove rather than fight with platform differences. The core features are already testing in the other tests.

include/fusilli/attributes/pointwise_attributes.h
rsuderman: I addressed the way described. It may also just make sense to do regular `enum` instead of `enum class`, then we can directly index. But either way this is pretty limited case.

rsuderman: We could even consider making the map a regular `std::array` of static size.

samples/pointwise/pointwise_binary_ops.cpp
```diff
@@ -151,5 +151,5 @@ TEST_CASE("Pointwise binary ops", "[pointwise][graph]") {
   // int32
   execute(handle, DataType::Int32, int(-50), int(13));
   // fp16
-  execute(handle, DataType::Half, half(-32.5f16), half(2.f16));
+  execute(handle, DataType::Half, half(-32.5f), half(2.f16));
```
sjain-stanford: suffix was dropped from first but not the second?

include/fusilli/support/dllib.h
sjain-stanford:
```suggestion
#if defined(FUSILLI_PLATFORM_WINDOWS)
```

include/fusilli/support/python_utils.h
```diff
@@ -84,14 +94,23 @@ inline std::optional<std::string> findIreeCompilerLib() {
   static std::optional<std::string> libPath;
   if (libPath.has_value())
     return libPath;
+#ifdef FUSILLI_PLATFORM_WINDOWS
```
sjain-stanford:
```suggestion
#if defined(FUSILLI_PLATFORM_WINDOWS)
```

sjain-stanford (approved):
Thanks for persevering through the many many fixes to get this Windows compliant. Looking forward to seeing the Windows CI green 🚀

include/fusilli/support/python_utils.h
rsuderman: Done.

include/fusilli/support/cache.h
```diff
@@ -101,6 +109,46 @@ class CacheFile {
...
+    // the GitHub Actions CI runners as well as for LIT tests that rely
+    // on dumping/reading intermediate compilation artifacts to/from disk.
+    const char *cacheDirEnv = std::getenv("FUSILLI_CACHE_DIR");
+    std::wstring cacheDir;
+    if (cacheDirEnv) {
+      std::string cacheDirStr(cacheDirEnv);
+      cacheDir = std::wstring(cacheDirStr.begin(), cacheDirStr.end());
+    }
+
+#ifdef FUSILLI_PLATFORM_WINDOWS
```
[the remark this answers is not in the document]
rsuderman: Done.

include/fusilli/support/process.h
```diff
@@ -43,8 +43,10 @@ inline std::optional<std::string> execCommand(const std::string &cmd) {
 
 #ifdef FUSILLI_PLATFORM_WINDOWS
   FILE *opened = _popen(cmd.c_str(), "r");
-#else
+#elif FUSILLI_PLATFORM_LINUX
```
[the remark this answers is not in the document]
rsuderman: Done.

PR #150, opened 2026-02-10
[Test] Add transient-memory-size checks to LIT tests

sjain-stanford:
Should help surface issues like these: https://github.com/iree-org/iree/issues/23408

## Summary
- Adds FileCheck assertions for `transient-memory-size` stats across all 42 LIT tests
- When `dispatch-count` is 1, verifies `transient-memory-size` is 0 (no intermediate buffers needed)
- When `dispatch-count` is greater than 1, verifies `transient-memory-size` is non-zero (using CHECK-NOT pattern)

## Test plan
- [x] All 131 tests pass locally
- [x] FileCheck ordering matches JSON output order (transient-memory-size before dispatch-count)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/benchmarks/driver.cpp
+++ b/benchmarks/driver.cpp
@@ -64,7 +64,7 @@ struct MatmulOptions {
 static ErrorObject benchmarkConvFprop(const ConvOptions &opts,
                                       DataType convIOType, int64_t iter,
                                       int64_t deviceId, bool dump) {
-#ifdef FUSILLI_ENABLE_AMDGPU
+#if defined(FUSILLI_ENABLE_AMDGPU)
   FUSILLI_ASSIGN_OR_RETURN(Handle handle,
                            Handle::create(Backend::AMDGPU, deviceId));
 #else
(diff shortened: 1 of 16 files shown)


tests/lit/test_conv_asm_emitter_nchw_kcrs_with_pad.cpp
```diff
@@ -58,7 +58,9 @@
 // LINALG-CHECK:      %[[OUT:.+]] = linalg.conv_2d_nchw_fchw {dilations = dense<1> : vector<2xi64>, strides = dense<1> : vector<2xi64>} ins(%[[BUF1PAD]], %[[BUF2]] : tensor<16x128x66x34xf32>, tensor<256x128x3x3xf32>) outs(%{{.+}} : tensor<16x256x64x32xf32>) -> tensor<16x256x64x32xf32>
 // LINALG-CHECK:      %{{.+}} = hal.tensor.alias wait(%{{.+}}) => %[[OUT]] : tensor<16x256x64x32xf32> to %[[ARG0]] : !hal.buffer_view
 //
+// AMDGPU-STATS-CHECK: "transient-memory-size": 0
 // AMDGPU-STATS-CHECK: "dispatch-count": 1
+// CPU-STATS-CHECK-NOT: "transient-memory-size": 0
 // CPU-STATS-CHECK: "dispatch-count": 2
```
sjain-stanford: When https://github.com/iree-org/fusilli/pull/132 lands, these `CHECK-NOT`s will be converted to `CHECK`s since we wouldn't be allocating for transients even for the multiple dispatch cases.

PR #155, opened 2026-02-12
Cleanup cmake files for non-dist level building

rsuderman:
We remove the build_dir include path from the fusilli library declaration. We do this because the includes are never copies to the build directory and therefore the path is unknown.

We make `hip` be a first order import on tests. This is to guarantee the the `hip_DIR` specified is used for imports over the hip includes in the iree third party repository.

### DIFF
--- a/plugins/hipdnn-plugin/src/fusilli_plugin.cpp
+++ b/plugins/hipdnn-plugin/src/fusilli_plugin.cpp
@@ -613,8 +613,8 @@ hipdnnPluginStatus_t hipdnnEnginePluginExecuteOpGraph(
     iree_hal_buffer_view_release(outBufferView);
   }
 
-  FUSILLI_PLUGIN_CHECK_ERROR(
-      executionContext->graph.execute(fusilliHandle, variantPack));
+  FUSILLI_PLUGIN_CHECK_ERROR(executionContext->graph.execute(
+      fusilliHandle, variantPack, /*workspace=*/nullptr));
 
   LOG_API_SUCCESS_AUTO("executed graph");
   return HIPDNN_PLUGIN_STATUS_SUCCESS;


CMakeLists.txt
```diff
@@ -168,7 +168,6 @@ file(GLOB_RECURSE FUSILLI_HEADERS
 target_sources(libfusilli INTERFACE
   FILE_SET fusilli_headers TYPE HEADERS # file sets require lowercase name
   BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/include
-            ${CMAKE_CURRENT_BINARY_DIR}/include
```
AaronStGeorge: We used to generate an include file based on some CMake variables here, we no longer do that so it's not necessary. I don't really understand why this would the empty dir would cause issues, but it seems reasonable to remove it 🤷‍♂️.

plugins/hipdnn-plugin/test/CMakeLists.txt
```diff
@@ -16,10 +16,10 @@ add_fusilli_plugin_test(
     test_fusilli_plugin_api.cpp
     utils.h
   DEPS
+    hip::host
     fusilli_plugin # Link plugin .so directly
     fusilli::fusilli
     GTest::gtest_main
-    hip::host
```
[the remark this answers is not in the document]
rsuderman: We should avoid assuming system deps. Its not impossible someone will want to build with a non-system version of hip which would again default to the iree hip deps.

CMakeLists.txt
rsuderman: :+1:

AaronStGeorge:
Looks like https://github.com/iree-org/fusilli/pull/132 broke the plugin build, I can take a look at that.

PR #159, opened 2026-02-13
Switch deprecated --iree-hip-target to --iree-rocm-target

sjain-stanford:
## Summary
- Migrate from deprecated `--iree-hip-target` compiler flag to `--iree-rocm-target`
- Rename helper function `getIreeHipTargetForAmdgpu` → `getIreeRocmTargetForAmdgpu` and update log messages
- Update corresponding tests and test tags (`hip-target` → `rocm-target`)


sjain-stanford:
> maybe remove the "Test plan" section from the description, since those three are tested by CI.

Thanks, I had the same thought - its something with the default CLAUDE prompt for creating PRs, so might need to customize it in the SKILL.md.

PR #161, opened 2026-02-16
[Fusilli,FusilliPlugin] Scalar value support

AaronStGeorge:
Adds support for scalar tensor value. Fusilli and the plugin will now support graphs of the form `alpha * A`, `alpha * A * B * beta`, etc.

### DIFF
--- a/plugins/hipdnn-plugin/test/integration/matmul/simple_scaled_matmul_accumulate.cpp
+++ b/plugins/hipdnn-plugin/test/integration/matmul/simple_scaled_matmul_accumulate.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 Advanced Micro Devices, Inc.
+// Copyright 2026 Advanced Micro Devices, Inc.
 //
 // Licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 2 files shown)


tests/lit/test_pointwise_asm_emitter_mul_scalar.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+// clang-format off
+//
+// Scalar IN_1 is NOT in the func.func signature — only result_ and arg0.
+// Scalar constant is emitted at graph level, before the node's permute ops.
+// TORCH-CHECK:   module @module {
+// TORCH-CHECK:     func.func @main(%result_: !torch.tensor<[2,3,128,128],f32>, %arg0: !torch.vtensor<[2,3,128,128],f32>) attributes {torch.assume_strict_symbolic_shapes} {
+// Graph-level scalar constant emission:
+// TORCH-CHECK:       %alpha = torch.vtensor.literal(dense<0x40000000> : tensor<1xf32>) : !torch.vtensor<[1],f32>
+// Node permute ops for IN_0 (non-scalar):
```
sjain-stanford: nit: can drop the explanatory comments inline here as they make it hard to read / separate out tests from comments :)

PR #182, opened 2026-02-22
Trim benchmark runner test commands

sjain-stanford:
## Summary
- Reduce `test_commands.txt` to one representative command per subcommand (conv, layernorm, matmul) plus one `[SKIP]` entry. The benchmark runner test validates the Python wrapper and driver infrastructure, not individual kernel correctness — redundant shape/dtype/flag variations added unnecessary JIT compilation overhead (~28 compilations down to ~8). For kernel variant specific benchmark tests, they're to be added via `add_fusilli_benchmark`. 
- Remove stale commented-out TODO benchmark from `CMakeLists.txt` (torch-mlir#4422).

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -249,14 +249,6 @@ add_fusilli_benchmark(
     --device 0 --iter 10 matmul -M 16 -N 32 -K 64 --a_type bf16 --b_type bf16 --out_type bf16
 )
 
-# TODO(llvm/torch-mlir#4422): Uncomment after torch-mlir issue is fixed.
-# add_fusilli_benchmark(
-#   NAME fusilli_benchmark_matmul_mixed
-#   DRIVER fusilli_benchmark_driver
-#   ARGS
-#     --device 0 --iter 10 matmul -M 16 -N 32 -K 64 --a_type f32 --b_type bf16 --out_type f32
-# )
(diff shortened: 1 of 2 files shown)


benchmarks/CMakeLists.txt
```diff
@@ -249,14 +249,6 @@ add_fusilli_benchmark(
     --device 0 --iter 10 matmul -M 16 -N 32 -K 64 --a_type bf16 --b_type bf16 --out_type bf16
 )
 
-# TODO(llvm/torch-mlir#4422): Uncomment after torch-mlir issue is fixed.
-# add_fusilli_benchmark(
-#   NAME fusilli_benchmark_matmul_mixed
-#   DRIVER fusilli_benchmark_driver
-#   ARGS
-#     --device 0 --iter 10 matmul -M 16 -N 32 -K 64 --a_type f32 --b_type bf16 --out_type f32
-# )
```
sjain-stanford: We decided not to support this after all to stay consistent with torch semantics (mixed precision matmuls are invalid in torch). Instead we have an elaborate check that only passes through the cases that are supported: https://sourcegraph.com/github.com/iree-org/fusilli@c97aceb5858cd7ee0e646d6223b2a63a5fef3c65/-/blob/include/fusilli/node/matmul_node.h?L155-L177

PR #183, opened 2026-02-23
[Fusilli] Updated initialization of weights/biases by non-trivial values in driver

a-sidorova:
Updated initialization of weights/biases by non-trivial values in driver to avoid possible compiler-side optimization removing redundant multiply/add operations.

### DIFF
--- a/benchmarks/driver.cpp
+++ b/benchmarks/driver.cpp
@@ -860,8 +860,8 @@ static CLI::App *registerLayerNormOptions(CLI::App &mainApp,
   layerNormApp->add_flag(
       "--elementwise_affine", layerNormOpts.elementwiseAffine,
       "Run in elementwise affine mode. This flag adds learnable per-element "
-      "affine parameters initialized to ones (for weights) and "
-      "zeros (for biases)");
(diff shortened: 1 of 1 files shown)


benchmarks/driver.cpp
```diff
@@ -567,13 +567,16 @@ static ErrorObject benchmarkLayerNormFwd(const LayerNormOptions &opts,
   std::unordered_map<std::shared_ptr<TensorAttr>, std::shared_ptr<Buffer>>
       variantPack = {{xT, xBuf}, {yT, yBuf}, {mT, mBuf}, {vT, vBuf}};
 
+  // If `elementwise_affine` is True, scale and bias buffers are
```
sjain-stanford: The help string could use an update:

```cpp
  // layerNormApp CLI flags:
  layerNormApp->add_flag(
      "--elementwise_affine", layerNormOpts.elementwiseAffine,
      "Run in elementwise affine mode. This flag adds learnable per-element "
      "affine parameters initialized to ones (for weights) and "
      "zeros (for biases)");
```

sjain-stanford (approved):

PR #186, opened 2026-02-24
Fix: Wrong handle in multi-stream handle test

sjain-stanford:
The multi-stream handle test failed flakily on a recent CI run on `main`: https://github.com/iree-org/fusilli/actions/runs/22334529677/job/64623964052

```
-------------------------------------------------------------------------------
Handle creation with stream and deviceId
  Create multiple handles with different HIP streams
-------------------------------------------------------------------------------
/workspace/tests/hip_tests/test_handle.cpp:92
...............................................................................

/workspace/tests/hip_tests/test_handle.cpp:136: FAILED:
  REQUIRE( readData == hostData )
with expansion:
  { 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.
  0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
  0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f }
  ==
  { 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.
  14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.
  14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f, 3.14f }

terminate called after throwing an instance of 'std::bad_variant_access'
  what():  std::get: wrong index for variant
/workspace/tests/hip_tests/test_handle.cpp:136: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```
- Fixed copy-paste bug in `test_handle.cpp` where the "Create multiple handles with different HIP streams" test was using `handle1` instead of `handle2` for the second buffer's `allocate` and `read` calls
- (I'm not sure if this is accurate) This could have caused a cross-stream race condition resulting in flaky D2H reads returning all zeros instead of the expected `3.14f` values

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,12 +271,6 @@ ASAN_OPTIONS=detect_leaks=1:halt_on_error=1 \
   ctest --test-dir build
 ```
 
-To customize UBSAN behavior at runtime, set the `UBSAN_OPTIONS` environment variable:
-```shell
-UBSAN_OPTIONS=print_stacktrace=1 \
-  ctest --test-dir build
-```
-
 Ensure `llvm-symbolizer` is in your `$PATH` (or set the `LLVM_SYMBOLIZER_PATH`
 environment variable) to get symbolized stack traces from sanitizers:
 ```shell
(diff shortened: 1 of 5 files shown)


AaronStGeorge:
> This could have caused a cross-stream race condition resulting in flaky D2H reads returning all zeros instead of the expected 3.14f values

The fix is definitely what we want, but I'm a little lost as to why it would cause a race condition. Everything should have been stream ordered on handle 1.

PR #188, opened 2026-02-25
Add CustomOpNode for embedding custom MLIR ops within Graph

IanWood1:
This PR gives users a reliable way to add operations that Fusilli doesn’t support yet, so unsupported ops can be prototyped and shipped without waiting for new built-in node types. It also allows a single MLIR string to work across multiple data types and tensor shapes, reducing duplication and making custom ops easier to maintain.  Custom ops are implemented via function definitions & calls because the function boundary gives a separate SSA scope and a clean interface to the graph.

### DIFF
--- a/include/fusilli/backend/runtime.h
+++ b/include/fusilli/backend/runtime.h
@@ -215,12 +215,15 @@ inline ErrorObject Graph::createVmContext(const Handle &handle,
 
   // Create HAL module and register it with the context.
   {
-    iree_hal_device_t *device = handle.getDevice();
+    iree_hal_device_group_t *deviceGroup = nullptr;
+    FUSILLI_CHECK_ERROR(iree_hal_device_group_create_from_device(
+        handle.getDevice(), allocator, &deviceGroup));
     iree_vm_module_t *halModule = nullptr;
     FUSILLI_CHECK_ERROR(iree_hal_module_create(
         handle.getInstance(), iree_hal_module_device_policy_default(),
-        /*device_count=*/1, &device, IREE_HAL_MODULE_FLAG_NONE,
+        deviceGroup, IREE_HAL_MODULE_FLAG_NONE,
         iree_hal_module_debug_sink_null(), allocator, &halModule));
+    iree_hal_device_group_release(deviceGroup);
     iree_status_t status = iree_vm_context_register_modules(
         vmContext_.get(), /*module_count=*/1, &halModule);
     iree_vm_module_release(halModule);
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
   "package-version": "0.0.1.dev",
-  "iree-version": "3.11.0rc20260217",
-  "therock-version": "7.12.0a20260217"
+  "iree-version": "3.11.0rc20260227",
+  "therock-version": "7.12.0a20260227"
 }


include/fusilli/attributes/custom_op_attributes.h
```diff
@@ -0,0 +1,70 @@
...
+  ///                                    %arg1: !torch.vtensor<[?],{IN1_DTYPE}>)
+  ///                                    -> !torch.vtensor<[?],{OUT0_DTYPE}> {
+  ///     ...
+  ///   }
+  ///
+  /// If no placeholders are present, the string is emitted verbatim.
+  CustomOpAttr &setMlir(std::string mlir) {
+    mlir_ = std::move(mlir);
+    return *this;
+  }
```
sjain-stanford: same (I think const ref is better here since this is usually called on lvalue that need not be copied, so the only copy is in the assignment)

include/fusilli/support/asm_emitter.h
```diff
@@ -270,6 +227,31 @@ inline std::string TensorAttr::getTensorTypeAsm(bool isValueTensor,
...
+         "TensorAttr::getDynamicTensorTypeAsm expects a valid data type");
+
+  std::ostringstream oss;
+  oss << (isValueTensor ? "!torch.vtensor<[" : "!torch.tensor<[");
+  for (size_t i = 0; i < getDim().size(); ++i) {
+    if (i > 0)
+      oss << ",";
+    oss << "?";
+  }
+  oss << "]," << kDataTypeToMlirTypeAsm.at(getDataType()) << ">";
```
sjain-stanford: Reuse the `interleave` pattern that `getTensorTypeAsm` does.

include/fusilli/support/asm_emitter.h
```diff
@@ -270,6 +227,31 @@ inline std::string TensorAttr::getTensorTypeAsm(bool isValueTensor,
   return oss.str();
 }
 
+// Emits a fully-dynamic ranked tensor type in MLIR assembly.
+// All dimensions are replaced with `?`.
+//
+// Example: TensorAttr with dim={4,8,16}, dataType=Half
+//   -> "!torch.vtensor<[?,?,?],f16>"
+inline std::string
+TensorAttr::getDynamicTensorTypeAsm(bool isValueTensor) const {
```
sjain-stanford: Did you consider extending `getTensorTypeAsm` to take a new argument (`isDynamic`) and reuse most of the common pieces?

include/fusilli/graph/graph.h
```diff
@@ -280,6 +282,43 @@ class Graph : public INode {
   std::shared_ptr<TensorAttr> reduction(const std::shared_ptr<TensorAttr> &x,
                                         ReductionAttr &attributes);
 
+  template <typename... Tensors>
+  std::vector<std::shared_ptr<TensorAttr>> customOp(CustomOpAttr &customOpAttr,
+                                                    Tensors &&...inputArgs) {
```
sjain-stanford: Can the body of this be moved out-of-line below, alongside other graph builder methods? The graph class is growing so this helps avoid endless scrolling to get an idea of the interface. And this will be better with the move to compiled lib as the definitions can just go in `.cc` to keep interface crisp.

include/fusilli/attributes/custom_op_attributes.h
```diff
@@ -0,0 +1,73 @@
...
+  /// Setters:
+  CustomOpAttr &setName(const std::string &name) {
+    name_ = name;
+    return *this;
+  }
+
+  /// Sets the MLIR function definition for this custom op.
+  ///
+  /// The string may contain placeholders that are resolved at emission time
+  /// by `CustomOpNode::resolveMlirPlaceholders()`:
```
sjain-stanford: Let's use plan `//` comments here and elsewhere instead of doxygen style.

tests/test_custom_op.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+  auto status = g.validate();
+  REQUIRE(isError(status));
+  REQUIRE(status.getCode() == ErrorCode::AttributeNotSet);
+  REQUIRE(status.getMessage() == "CustomOp input 0 is null");
+}
+
+// Duplicate inputs (same tensor in multiple slots) must produce valid ASM
+// with unique SSA names — no redefinition errors.
+TEST_CASE("CustomOp duplicate input produces valid ASM", "[custom_op]") {
```
sjain-stanford: Consider promoting this to a dedicated LIT test.

include/fusilli/node/node.h
```diff
@@ -66,6 +67,15 @@ class INode {
...
   virtual std::string emitNodePostAsm() const { return ""; };
+  virtual std::string emitModuleScopeAsm() const { return ""; };
+
+  // Recursively collect module-scope ASM declarations from this node
+  // and all sub-nodes (e.g., custom op function definitions).
+  void collectModuleScopeAsm(std::ostringstream &oss) const {
+    oss << emitModuleScopeAsm();
+    for (const auto &subNode : subNodes_)
+      subNode->collectModuleScopeAsm(oss);
+  }
```
sjain-stanford: Would be nice to add a unit test for `collectModuleScopeAsm` on a graph that uses this recursive nesting of subnodes if possible.

tests/test_custom_op.cpp
```diff
@@ -0,0 +1,336 @@
...
+  outs[0]
+      ->setDim({4})
+      .setStride({1})
+      .setDataType(DataType::Float)
+      .setOutput(true);
+
+  auto status = g.validate();
+  REQUIRE(isError(status));
+  REQUIRE(status.getCode() == ErrorCode::InvalidAttribute);
+  REQUIRE(status.getMessage() == "CustomOp input 0 is scalar (not supported)");
```
sjain-stanford: Add test for output scalar as well.

include/fusilli/node/custom_op_node.h
```diff
@@ -0,0 +1,137 @@
...
+                              "CustomOp input " + std::to_string(i) +
+                                  " is null");
+      FUSILLI_RETURN_ERROR_IF(
+          inputs[i]->isScalar(), ErrorCode::InvalidAttribute,
+          "CustomOp input " + std::to_string(i) + " is scalar (not supported)");
+    }
+    for (size_t i = 0; i < outputs.size(); ++i) {
+      FUSILLI_RETURN_ERROR_IF(!outputs[i], ErrorCode::AttributeNotSet,
+                              "CustomOp output " + std::to_string(i) +
+                                  " is null");
```
sjain-stanford: If this is a valid scenario, please add unit test. If not, the check may be moot.

include/fusilli/graph/graph.h
```diff
@@ -280,6 +282,10 @@ class Graph : public INode {
   std::shared_ptr<TensorAttr> reduction(const std::shared_ptr<TensorAttr> &x,
                                         ReductionAttr &attributes);
 
+  template <typename... Tensors>
+  std::vector<std::shared_ptr<TensorAttr>> customOp(CustomOpAttr &customOpAttr,
+                                                    Tensors &&...inputArgs);
```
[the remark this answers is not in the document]
IanWood1: Good point, I'll change this to use `std::vector`. Also, I think "pass-by-value, then move" should be fine here since the vector is almost always a rvalue.

include/fusilli/node/node.h
IanWood1: We don't currently have any nodes that do nesting so I added a test class in [`1c10901` (this PR)](https://github.com/iree-org/fusilli/pull/188/commits/1c10901658e112d598f636768a63d43b555cfa9c). Is this what you wanted?

include/fusilli/node/custom_op_node.h
```diff
@@ -0,0 +1,134 @@
...
+                              "CustomOp output " + std::to_string(i) +
+                                  " is scalar (not supported)");
+    }
+
+    return ok();
+  }
+
+  // Resolves `{FUNC_NAME}`, `{IN<i>_DTYPE}`, and `{OUT<i>_DTYPE}` placeholders
+  // in the MLIR template using the node's name and tensor data types.
+  std::string resolveMlirPlaceholders() const {
```
sjain-stanford: Is it worth having a validation check that ensures no placeholders were left unresolved? Or is that guaranteed to be the case.

include/fusilli/support/asm_emitter.h
```diff
@@ -360,15 +367,27 @@ inline std::string Graph::getResultNamesAndTypesAsm() const {
...
-                                   getResultNamesAndTypesAsm(), // {0}
-                                   getOperandNamesAndTypesAsm() // {1}
-  );
+  // The func.func schema uses string concatenation (not std::format) to
+  // avoid issues with braces in module-scope MLIR content.
+  output += "  func.func @main(";
+  output += getResultNamesAndTypesAsm();
+  output += ", ";
+  output += getOperandNamesAndTypesAsm();
+  output += ") attributes {torch.assume_strict_symbolic_shapes} {\n  ";
```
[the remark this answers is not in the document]
sjain-stanford: Don't block on this though, best to fix indents in a separate one to keep concerns separate.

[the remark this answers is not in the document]
IanWood1: Whoops I forgot to send the comment. Yep, these are unrelated and I'll fix it in a follow up.

include/fusilli/node/custom_op_node.h
IanWood1: I thought about this, but I don't think there is a good way to do this without being overly broad since the user's MLIR may contain braces for regions or attr dicts. If one slips through the iree-compile error is fairly useful. I'm going to merge this but let me know if you have an idea to validate and I can add it as a follow up.

sjain-stanford: Yeah let's cross the bridge if this becomes an issue.

PR #198, opened 2026-02-26
Add nightly dependency bump workflow

sjain-stanford:
## Summary
- Add `nightly-bump.yml` scheduled GitHub Actions workflow that runs daily at 10:00 UTC
- Discovers latest IREE and TheRock nightly versions (with pip wheel availability fallback)
- Updates `version.json` and creates a PR via `peter-evans/create-pull-request`
- Supports manual `workflow_dispatch` with optional version overrides for testing
- No-op detection skips PR creation when versions haven't changed
- Fork guard prevents scheduled runs on forks

Depends on #199 (decouples version bumps from Docker image rebuilds, adds `therock-version` to `version.json`).

### DIFF
--- a/.github/workflows/nightly-bump.yml
+++ b/.github/workflows/nightly-bump.yml
@@ -1,0 +1,173 @@
+# Copyright 2026 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: Nightly Dependency Bump
+
+on:
+  workflow_dispatch:
+    inputs:
+      iree_version:
+        description: "IREE version override (e.g., 3.11.0rc20260217). Skips auto-discovery."
+        required: false
+      therock_version:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-bump.yml
```diff
@@ -0,0 +1,152 @@
...
+                 --dry-run 2>/dev/null; then
+            echo "::warning::Latest IREE wheel ${LATEST} not available, falling back to ${FALLBACK}"
+            CHOSEN=${FALLBACK}
+          else
+            echo "::error::No IREE wheel available for ${LATEST} or ${FALLBACK}"
+            exit 1
+          fi
+
+          echo "Chosen IREE version: ${CHOSEN}"
+          echo "version=${CHOSEN}" >> $GITHUB_OUTPUT
```
ScottTodd: I'd suggest pulling this logic into a script like https://github.com/iree-org/iree-turbine/blob/main/build_tools/update_iree_requirement_pins.py , used by a workflow like https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/update_iree_requirement_pins.yml. This sort of bash logic in a workflow file is very difficult to debug and extend.

PR #241, opened 2026-03-12
Add built-in SDPA forward op

sjain-stanford:
## Summary

- Promotes SDPA forward from a custom op (#227) to a first-class built-in op with `SdpaAttr`, `SdpaNode`, and native ASM emission via `torch.aten.scaled_dot_product_attention`
- New API: `graph->sdpa(Q, K, V, mask, attr)` with validation (rank 4, head matching, GQA, mask/causal exclusion, dropout range) and automatic output shape inference
- Migrates benchmark driver to use the built-in `graph.sdpa()` API instead of the custom op path
- Existing custom op SDPA samples preserved under `samples/custom_op/sdpa/` for reference; `samples/sdpa/` rewritten to use the built-in API
- Custom op SDPA templates updated to use static `{INx_TYPE}` placeholders (matching static MLIR support from #273)


include/fusilli/node/sdpa_node.h
```diff
@@ -0,0 +1,206 @@
...
+        qDim[0] != kDim[0] || qDim[0] != vDim[0], ErrorCode::InvalidAttribute,
+        "SDPA input tensors Q, K, V must have matching batch dimension");
+
+    // Head dimension must match across Q and K.
+    FUSILLI_RETURN_ERROR_IF(
+        qDim[3] != kDim[3], ErrorCode::InvalidAttribute,
+        "SDPA input tensors Q and K must have matching head_dim");
+
+    // K and V must have matching sequence length and heads.
+    // TODO(#280): Relax the h_k == h_v constraint once IREE fix is in.
```
rsuderman, line 109: Can we make this a link to the issue? Guarantees we can identify it without inferring repository.

include/fusilli/support/asm_emitter.h
```diff
@@ -1926,6 +1927,154 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+
+  // Permute inputs.
+  std::string permuteQ = getLayoutConversionOpsAsm(sdpaAttr.getQ(), "permute_Q",
+                                                   suffix, /*isInput=*/true);
+  std::string permuteK = getLayoutConversionOpsAsm(sdpaAttr.getK(), "permute_K",
+                                                   suffix, /*isInput=*/true);
+  std::string permuteV = getLayoutConversionOpsAsm(sdpaAttr.getV(), "permute_V",
+                                                   suffix, /*isInput=*/true);
+
+  std::string permuteMask;
```
rsuderman: Just declare `std::String mask` and set in the right scenario. We should not need separate variables to track.

include/fusilli/support/asm_emitter.h
```diff
@@ -1926,6 +1927,154 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+                     permuteQ,                                     // {0}
+                     permuteK,                                     // {1}
+                     permuteV,                                     // {2}
+                     permuteMask.empty() ? noneMask : permuteMask, // {3}
+                     getDropoutOpsAsm(),                           // {4}
+                     getIsCausalOpsAsm(),                          // {5}
+                     getScaleOpsAsm(),                             // {6}
+                     getEnableGqaOpsAsm(),                         // {7}
+                     getResultNamesAsm(),                          // {8}
+                     getOperandNamesAsm() + ", %dropout_" + suffix +
```
rsuderman: Lets not generate this inlined. Its a look string format, better to land somewhere above.

samples/sdpa_utils.cpp
```diff
@@ -0,0 +1,322 @@
...
+///
+/// When hasAttnMask is false: 3 tensor inputs (Q=IN0, K=IN1, V=IN2).
+/// When hasAttnMask is true:  4 tensor inputs (Q=IN0, K=IN1, V=IN2, mask=IN3).
+static std::string buildSdpaMlir(bool hasAttnMask = false,
+                                 float dropoutP = 0.0f, bool isCausal = false,
+                                 std::optional<float> scale = std::nullopt,
+                                 bool enableGqa = false) {
+  // Compute raw values for scalar placeholders.
+  std::string dropoutStr = std::format("{:e}", dropoutP);
+  std::string isCausalStr = isCausal ? "true" : "false";
```
rsuderman, line 33: Can we make some IR helpers. This feels quite repeatable.

samples/custom_op/sdpa/sdpa_utils.h
```diff
@@ -0,0 +1,322 @@
+// Copyright 2026 Advanced Micro Devices, Inc.
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// NOLINTNEXTLINE(llvm-header-guard)
+#ifndef FUSILLI_SAMPLES_SDPA_SDPA_UTILS_H
```
IanWood1: Conflicts with [samples/sdpa/sdpa_utils.h](https://github.com/iree-org/fusilli/pull/241/changes#diff-b69533bf4f6cf4134795edd1ea5dbd1c7be2fe8f66d8acf9de5a376cff9e509e)

```suggestion
#ifndef FUSILLI_SAMPLES_CUSTOM_OP_SDPA_SDPA_UTILS_H
```

samples/custom_op/sdpa/sdpa_utils.h
```diff
@@ -0,0 +1,322 @@
...
+
+/// Builds the MLIR template for torch.aten.scaled_dot_product_attention.
+///
+/// Selects the appropriate R-string template (with/without attn_mask) and
+/// resolves scalar placeholders. Standard CustomOp dtype/name placeholders
+/// are left for CustomOpNode to resolve at emission time.
+///
+/// When hasAttnMask is false: 3 tensor inputs (Q=IN0, K=IN1, V=IN2).
+/// When hasAttnMask is true:  4 tensor inputs (Q=IN0, K=IN1, V=IN2, mask=IN3).
+static std::string buildSdpaMlir(bool hasAttnMask = false,
```
IanWood1: Should be `inline`: 

```suggestion
inline std::string buildSdpaMlir(bool hasAttnMask = false,
```


Ignore if this gets moved to a cpp file

samples/custom_op/sdpa/sdpa_utils.h
```diff
@@ -0,0 +1,322 @@
...
+                                            scaleTypeStr,  // {3} SCALE_TYPE
+                                            enableGqaStr   // {4} ENABLE_GQA
+                                            ));
+}
+
+// CPU reference implementation of scaled dot-product attention.
+// Computes SDPA in float precision for numerical verification against the GPU.
+// Layout: [batch, heads, seq_len, head_dim] contiguous.
+static std::vector<float>
+referenceSdpa(float qVal, float kVal, float vVal, float maskVal, int64_t batch,
```
IanWood1: Should be `inline`:

```suggestion
inline std::vector<float>
referenceSdpa(float qVal, float kVal, float vVal, float maskVal, int64_t batch,
```

samples/sdpa/sdpa_utils.h
IanWood1: This same function appears in samples/custom_op/sdpa/sdpa_utils.h

samples/custom_op/sdpa/sdpa_utils.h
```diff
@@ -0,0 +1,322 @@
+// Copyright 2026 Advanced Micro Devices, Inc.
```
IanWood1, line 1: Could we add a `sdpa_utils.cpp` and move the implementations there? Same for the other sdpa_utils.cpp.

rsuderman (changes requested):

include/fusilli/node/sdpa_node.h
```diff
@@ -0,0 +1,206 @@
...
+    constexpr size_t kRequiredRank = 4;
+    FUSILLI_RETURN_ERROR_IF(
+        qT->getDim().size() != kRequiredRank, ErrorCode::InvalidAttribute,
+        "SDPA input tensor Q must be rank 4 [batch, heads, seq_len, head_dim]");
+    FUSILLI_RETURN_ERROR_IF(
+        kT->getDim().size() != kRequiredRank, ErrorCode::InvalidAttribute,
+        "SDPA input tensor K must be rank 4 [batch, heads, seq_len, head_dim]");
+    FUSILLI_RETURN_ERROR_IF(
+        vT->getDim().size() != kRequiredRank, ErrorCode::InvalidAttribute,
+        "SDPA input tensor V must be rank 4 [batch, heads, seq_len, head_dim]");
```
IanWood1: Nit: it might be worth differentiating source/target seq len and q vs kv head count.

include/fusilli/node/sdpa_node.h
```diff
@@ -0,0 +1,206 @@
...
+    FUSILLI_RETURN_ERROR_IF(
+        maskT && sdpaAttr.getIsCausal(), ErrorCode::InvalidAttribute,
+        "SDPA attention mask and is_causal are mutually exclusive");
+
+    // Mask rank check.
+    if (maskT) {
+      FUSILLI_RETURN_ERROR_IF(maskT->getDim().size() != kRequiredRank,
+                              ErrorCode::InvalidAttribute,
+                              "SDPA attention mask must be rank 4");
+    }
```
IanWood1: Do we also want to check the mask dims against the other inputs for compatible batch & seq len?

include/fusilli/node/sdpa_node.h
sjain-stanford: We use `TODO(#num)` for issues in the same repo and `TODO(org/repo#num)` for cross repo issues. I believe it's the same convention as [IREE](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/iree-org/iree%24+TODO%28%23&patternType=keyword&sm=0) / [LLVM](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/llvm/llvm-project%24+TODO%28%23&patternType=keyword&sm=0).

include/fusilli/support/asm_emitter.h
sjain-stanford: Claude being claude. Fixed.

include/fusilli/support/asm_emitter.h
sjain-stanford: Fixed.

samples/custom_op/sdpa/sdpa_utils.h
sjain-stanford: fixed

samples/custom_op/sdpa/sdpa_utils.h
sjain-stanford: Fixed.

samples/sdpa_utils.cpp
sjain-stanford: Added 4 IR constant helpers in `asm_emitter.h` (lines 131-158):
  • torchNoneAsm(name, suffix) — %name_suffix = torch.constant.none
  • torchBoolAsm(name, suffix, val) — %name_suffix = torch.constant.bool true/false
  • torchFloatAsm(name, suffix, val) — %name_suffix = torch.constant.float 1.0e+00
  • torchIntAsm(name, suffix, val) — %name_suffix = torch.constant.int 42

  Applied across 12 call sites replacing raw std::format calls:
  • SdpaNode: getDropoutOpsAsm, getIsCausalOpsAsm, getScaleOpsAsm, getEnableGqaOpsAsm,
    emitNodePreAsm (none_mask)
  • Conv nodes: 3x getGroupCountOpsAsm (FProp, WGrad, DGrad)
  • BatchNormNode: getNoneOrEmpty lambda
  • LayerNormNode: scale/bias none fallbacks
  • RMSNormNode: scale none fallback

  The sdpa_utils.h sample wasn't changed — it uses a template approach where
  torch.constant.* ops are baked into the R-string and only raw values get plugged in, so
  the full SSA helpers don't apply there.

samples/sdpa/sdpa_utils.h
sjain-stanford: I've consolidated the two into a single `libsdpautils` with implementations in .cpp.

samples/custom_op/sdpa/sdpa_utils.h
sjain-stanford: I've consolidated the two into a single `libsdpautils` with implementations in .cpp.

include/fusilli/node/sdpa_node.h
sjain-stanford: Per PT [docs](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html), mask needs to be broadcastable to attention weights - (N,...,L,S) where N - batch size, L - target sequence length, S - source sequence length. Added checks for dim to be 1 or exact match (https://github.com/iree-org/fusilli/pull/241/commits/72461b99759f069f4e9f98491c4534b81b631171). LMK if I missed something.

rsuderman (approved):

PR #248, opened 2026-03-13
feat: Add RMS normalization (rmsnorm) op support, following layernorm

MaheshRavishankar:
Add RmsNormNode following the same pattern as the existing LayerNormNode:
- RmsnormAttr with X, SCALE (optional), EPSILON inputs and Y, INV_RMS outputs
- RmsNormNode with validation, property inference, and ASM emission
- Inference mode emits torch.aten.rms_norm (supported by IREE)
- Training mode emits torch.aten.native_rms_norm (not yet supported by IREE)

Tests:
- Unit tests for attributes and node validation/inference
- Lit emitter tests for inference (NCHW, NHWC+scale) with full TORCH-CHECK and compilation stats verification
- Lit emitter tests for training (NCHW, NHWC+scale) with TORCH-CHECK only; iree-opt roundtrip and stats checks are disabled until IREE adds torch.aten.native_rms_norm support
- End-to-end inference samples (NCHW, NCHW+scale, NHWC+scale)
- Training end-to-end samples are excluded pending IREE compiler support

### DIFF
--- a/include/fusilli/node/norm_utils.h
+++ b/include/fusilli/node/norm_utils.h
@@ -24,8 +24,8 @@ namespace fusilli::norm_utils {
 // Returns the shape over which normalization is applied:
 // the input tensor's shape excluding the batch dimension,
 // as normalization is computed independently for each sample in the batch.
-inline std::vector<int64_t>
-getNormalizedShape(const std::vector<int64_t> &xDim, size_t batchDim = 0) {
+inline std::vector<int64_t> getNormalizedShape(const std::vector<int64_t> &xDim,
+                                               size_t batchDim = 0) {
   std::vector<int64_t> shape;
   shape.reserve(xDim.size() - 1);
   for (size_t i = 0; i < xDim.size(); ++i) {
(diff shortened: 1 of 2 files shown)


MaheshRavishankar:
This was mostly done to be a mirror of RMS Norm. The iree-compiler does not support lowering of rms norm operation in torch for training. That needs to be done as a follow up when torch-mlir support is added. So there are no end to end tests for that.

One question though, both Layernorm (and as a result the RMSNorm implementation here) dont seem to support broadcasting behavior for the Scale. So the Scale tensor is expected to be `[1, C, H, W]` when the input is `[N, C, H, W]`, when it could just be `[C]`. The same is true for the `INV_RMS` that is output for the training variant. It expects a shape of `[N, 1, 1, 1]` instead of `[N]`. Is that expected, or do we fix that as a follow up for both LayerNorm and RMSNorm?

include/fusilli/attributes/rmsnorm_attributes.h
```diff
@@ -0,0 +1,69 @@
...
+#include <cstdint>
+#include <memory>
+#include <unordered_map>
+
+namespace fusilli {
+
+class RmsnormAttr : public AttributesCRTP<RmsnormAttr> {
+public:
+  // Names for Tensor Inputs and Outputs.
+  enum class InputNames : uint8_t { X, SCALE, EPSILON };
```
IanWood1: Does this PR omit `BIAS` because the torch op doesn't support it?

PR #262, opened 2026-03-23
[NFC] Add error propagation to ASM emitter boundary methods

MaheshRavishankar:
Replace assertions with proper ErrorOr<std::string> error propagation at the ASM emitter boundary (emitNodePreAsm, emitNodePostAsm, emitModuleScopeAsm). This allows errors to propagate gracefully back to the user instead of crashing via assert in release builds.

Changes:
- INode virtual methods return ErrorOr<std::string> instead of std::string
- emitAsmSubtree and collectModuleScopeAsm return ErrorObject
- Graph::emitAsm propagates errors from the emitter tree
- Replace assert with FUSILLI_RETURN_ERROR_IF for:
  - RmsNorm training mode (not yet supported)
  - Unsupported pointwise mode
  - Unsupported reduction mode
- Internal helpers (getTensorTypeAsm, getValueNameAsm, etc.) retain asserts as defensive checks since validation guarantees correctness

### DIFF
--- a/include/fusilli/graph/graph.h
+++ b/include/fusilli/graph/graph.h
@@ -306,7 +306,7 @@ public:
         !isValidated_, ErrorCode::NotValidated,
         "Graph must be validated before emitting MLIR assembly");
     std::ostringstream oss;
-    emitAsmSubtree(oss);
+    FUSILLI_CHECK_ERROR(emitAsmSubtree(oss));
     FUSILLI_LOG_ENDL(oss.str());
     return ok(oss.str());
   }
@@ -578,8 +578,8 @@ private:
   ErrorObject postValidateNode() const override final { return ok(); }
 
   // MLIR assembly emitter helper methods.
-  std::string emitNodePreAsm() const override final;
-  std::string emitNodePostAsm() const override final;
+  ErrorOr<std::string> emitNodePreAsm() const override final;
+  ErrorOr<std::string> emitNodePostAsm() const override final;
   std::string getOperandNamesAndTypesAsm() const;
   std::string getResultNamesAndTypesAsm() const;
 
--- a/include/fusilli/node/conv_node.h
+++ b/include/fusilli/node/conv_node.h
@@ -67,7 +67,7 @@ public:
       : NodeCRTP(ctx), convFPropAttr(std::move(attr)) {}
 
   // MLIR assembly emitter helper methods.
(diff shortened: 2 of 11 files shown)


MaheshRavishankar:
> @MaheshRavishankar, @IanWood1 Sorry I couldn't get to this while I was on leave.
> 
> This breaks the assumption that the asm emitter methods never error (if they do that's a failure on the graph validation logic and needs an update). I'd like to better understand the situation that spawned this and see if there's another way to address the underlying issue (why not use debug builds for dev work)? Can you share the repro / error message and what change triggered it?
> 
> Usually if these asserts are hit (shouldn't in release builds as that code is unreachable) it is highly likely that something is off in a fresh development branch, which ideally uses debug builds. This change is also very non-uniform - doesn't hold to rest of asm emitters, so I push to revert and then revisit if there's a hidden need I'm not seeing rn.

I think it is a matter of balance.  I can discuss more, but for example one iteration of this change removed all the asserts in the emitter and propagated the error to the user. That was a very invasive change and there I think just having asserts is OK. but at a "top level" it might be better to catch errors and return it. Here for example, rms norm for training is not supported through Torch-MLIR (there is no op that exists). The options were to remove support for RMS Norm training support from Fusilli completely, but really that did not make sense to me. So I took an intermediate step where RMS Norm training is supported in the Fusilli front-end, but the emitter errors out. I did add a post verification check that throws an error but there is a bit of an "action at a distance". For whatever reason if the post-verification changed, then the only guard rail left is an assert (which is not really a guard rail). So a simple error handling at that level seemed to provide some more guard rail without it being too invasive. I felt that trade-off is OK.

sjain-stanford:
> Here for example, rms norm for training is not supported through Torch-MLIR (there is no op that exists). The options were to remove support for RMS Norm training support from Fusilli completely, but really that did not make sense to me. So I took an intermediate step where RMS Norm training is supported in the Fusilli front-end, but the emitter errors out. I did add a post verification check that throws an error but there is a bit of an "action at a distance". 

The right thing to do for this should be a pre-validation check that fails for training RMS norm. We should just add that (if it's not already there) and add a unit test that ensures it stays that way until a time this is supported by Fusilli. The ASM emitter shouldn't error out for such cases. I know where you're coming from but this is kind of a non-ideal compromise for something that is baked deep in the design of the ASM emitter (which shouldn't take on the role of validating fusilli graphs).

PR #263, opened 2026-03-24
Add SDPA forward benchmark driver support

keshavvinayak01:
Adds `sdpa` subcommand to the benchmark driver using `graph.customOp()` with MLIR templates for `torch.aten.scaled_dot_product_attention`. Supports causal masks, attention masks, and f16/bf16/f32 data types.

Co-Authored-By: Claude Opus 4.6 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,7 +156,7 @@ if(IREE_SOURCE_DIR)
   add_subdirectory(${IREE_SOURCE_DIR} fusilli_iree SYSTEM EXCLUDE_FROM_ALL)
 else()
   message(STATUS "Fetching IREE sources from tag ${IREE_GIT_TAG}")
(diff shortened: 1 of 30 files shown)


benchmarks/CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ target_link_libraries(fusilli_benchmark_driver PRIVATE
   libutils
   CLI11::CLI11
 )
+target_include_directories(fusilli_benchmark_driver PRIVATE
+  ${CMAKE_SOURCE_DIR}/samples/sdpa
+)
```
IanWood1: We probably don't want benchmarks to depend on samples. This might be fine since the custom op implementation of sdpa should be replaced soon but could you add a TODO/warning about this.

Issue #233, opened 2026-03-06
Custom op & shape handling


sjain-stanford:
Custom ops are opaque blobs of MLIR asm that Fusilli gets from the user and feeds through to the IREE compiler as is. To make authoring these custom ops easy, certain things (like shapes, types) maybe left unmaterialized for Fusilli to materialize. 

For shapes, there are a few options at our disposal:

- Option 1: Custom MLIR with full static shapes
- Option 2: Custom MLIR with full dynamic shapes + user provided shape replacement dict
- Option 3: Custom MLIR with full dynamic shapes + torch shape inference

We currently support option 3 in Fusilli.

Option 2 (with user provided shape dict) is not desirable for two reasons:
a) If user goes the extra mile to specify shape replacements, why not just do the shape replacement themselves and give Fusilli a static MLIR (option 1) to begin with? 
b) Verifying that the user provided shape replacements match the actual shapes from TensorAttr at runtime

Option 1 is doable but won't work OOTB today due to the dynamic shape casts. But probably easy to fix by taking a bool attr on the custom op `isDynamic` or something.

PR #273, opened 2026-03-26
Add shape placeholders and static types for custom op MLIR templates

IanWood1:
Previously, the always generated dynamic shapes (`[?]`) in `func.call` signatures and inserted `torch.tensor_static_info_cast` ops to bridge the graph's static types and the call's dynamic types. This meant:
  1. Users couldn't write MLIR templates with hardcoded static shapes.
  2. The compiler had to re-infer shapes that fusilli already knew (fragile).

#272 removed the `isStatic` flag (which was the previous attempt at solving this, but was difficult to plumb through hipdnn serialization). This PR takes the approach discussed in #233: extend the placeholder system so the MLIR string is the only thing that needs to be stored in  `CustomOpAttr`.

## Changes:
- Extend the custom op placeholder system with `{IN0_TYPE}`, `{OUT0_TYPE}`, and `{IN0_DIM0}` placeholders that resolve to static tensor types and individual dimensions
- Switch the emitter to generate fully static types in `func.call`
- Users can now write MLIR templates with hardcoded static types (no placeholders needed if dimensions are known)

### DIFF
--- a/.github/workflows/bump-deps.yml
+++ b/.github/workflows/bump-deps.yml
@@ -9,9 +9,14 @@ name: Bump Deps (nightly)
 on:
   workflow_dispatch:
     inputs:
-      branch-name:
-        type: string
-        description: "Branch name override (default: bump-deps/YYYY-MM-DD)"
+      dep:
+        type: choice
+        description: "Which dependency to bump"
+        default: both
+        options:
+          - both
+          - iree
+          - therock
   schedule:
     # Weekdays at 12:00 UTC (04:00 PST / 05:00 PDT)
     - cron: "0 12 * * 1-5"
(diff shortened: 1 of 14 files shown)


include/fusilli/attributes/custom_op_attributes.h
```diff
@@ -42,14 +42,22 @@ class CustomOpAttr {
...
+  // All indices are 0-based and generalize to any input/output/dimension count
+  // (e.g., {IN2_TYPE}, {OUT1_DIM3}).
   //
   // Example:
-  //   func.func private @{FUNC_NAME}(%arg0: !torch.vtensor<[?],{IN0_DTYPE}>,
-  //                                    %arg1: !torch.vtensor<[?],{IN1_DTYPE}>)
-  //                                    -> !torch.vtensor<[?],{OUT0_DTYPE}> {
+  //   func.func private @{FUNC_NAME}(%arg0: {IN0_TYPE},
+  //                                    %arg1: {IN1_TYPE})
+  //                                    -> {OUT0_TYPE} {
```
sjain-stanford: I think there's a distinction between some of these. 

FUNC_NAME: mandatory - can't be hardcoded
*_DTYPE: mandatory (I think)
*_TYPE: optional

Would be useful to document this explicitly as this can be confusing to the user.

sjain-stanford (commented):
It is not clear if we do support option 3 or not. I see the shape casts are removed so I assume not - can you confirm and document in the issue #233 ? If not, we should remove the shape inference pass from the backend compile flags.



> Option 1: Custom MLIR with full static shapes
Option 2: Custom MLIR with full dynamic shapes + user provided shape replacement dict
Option 3: Custom MLIR with full dynamic shapes + torch shape inference
Option 4: Custom MLIR with shape placeholders + TensorAttr based replacements

PR #311, opened 2026-04-03
Add ELU_FWD pointwise op support

rsuderman:


samples/pointwise/pointwise_unary_ops.cpp
```diff
@@ -72,6 +72,7 @@ TEST_CASE("Pointwise unary ops", "[pointwise][graph]") {
   const auto mode = GENERATE(
       PointwiseAttr::Mode::ABS,
       PointwiseAttr::Mode::CEIL,
+      PointwiseAttr::Mode::ELU_FWD,
```
sjain-stanford, line 82: I think adding here isn't sufficient and this might need to be added to `supportsFloat` to really be tested. It's annoying but we might want to have a check that ensures ops are in atleast one of the two - to avoid them being silently ignored.

rsuderman: It should be running. I added the verification that its executing atleast one test part

sjain-stanford (approved):

PR #318, opened 2026-04-07
Flake test: drop --extra-verbose to reduce log noise

IanWood1:
`--output-on-failure` (always on in test.sh) already prints full output for any test that fails. `--extra-verbose` prints full output for every passing run too.


sjain-stanford (approved):
Should we remove this from regular CI as well?

PR #321, opened 2026-04-07
feat: Add built-in SDPA forward op using online_attention

Groverkss:
Modification to https://github.com/iree-org/fusilli/pull/241 to use online_attention.


sjain-stanford:
Discussed offline (https://discord.com/channels/689900678990135345/1491175878141542510/1491574537156821114) and noting here for the record: The POR is to target flex attention from fusilli instead and rely on torch.flex_attention -> linalg_ext.online_attention lowering to handle the conversions to a form IREE can handle.

PR #325, opened 2026-04-08
Add AMAX, AVG, NORM1, NORM2, MUL, MUL_NO_ZEROS reduction modes

rsuderman:
Enable the remaining cuDNN reduction modes in ReductionAttr and add the corresponding MLIR schemas to the asm emitter:

- NORM1 lowers to abs + sum.dim_IntList.
- AMAX lowers to abs + amax.
- AVG lowers to mean.dim (float dtypes only — torch.aten.mean.dim is not defined on integer tensors, so the sample skips int32 for AVG).
- NORM2 lowers to mul + sum.dim_IntList + sqrt.
- MUL lowers directly to torch.prims.prod.
- MUL_NO_ZEROS uses aten.ne.Scalar to build an i1 mask, then aten.where.ScalarOther to substitute 1 for zero entries before feeding the result to torch.prims.prod, so zero inputs are excluded from the product.

Extend samples/reduction/reduction_ops.cpp to exercise every new mode. Input data is built by a per-mode generateReductionInputData helper so MUL/MUL_NO_ZEROS get a non-trivial pattern (mostly 1s with a 2 and a 3, plus injected zeros for MUL_NO_ZEROS) that stays in range for fp16/int32, and the expected value is computed by the existing reference reduction loop rather than hardcoded.

Add lit tests for each new mode under tests/lit/ and register them in tests/CMakeLists.txt.

### DIFF
--- a/include/fusilli/attributes/reduction_attributes.h
+++ b/include/fusilli/attributes/reduction_attributes.h
@@ -26,6 +26,7 @@ namespace fusilli {
 
 #define FUSILLI_REDUCTION_MODES(OP)                                            \
   OP(NOT_SET)                                                                  \
+  OP(ADD)                                                                      \
   OP(SUM)                                                                      \
   OP(MUL)                                                                      \
   OP(MIN)                                                                      \
--- a/include/fusilli/attributes/reduction_attributes.h
+++ b/include/fusilli/attributes/reduction_attributes.h
@@ -27,15 +27,14 @@ namespace fusilli {
 #define FUSILLI_REDUCTION_MODES(OP)                                            \
   OP(NOT_SET)                                                                  \
   OP(SUM)                                                                      \
-  /* OP(ADD) */                                                                \
-  /* OP(MUL) */                                                                \
+  OP(MUL)                                                                      \
   OP(MIN)                                                                      \
   OP(MAX)                                                                      \
-  /* OP(AMAX) */                                                               \
-  /* OP(AVG) */                                                                \
-  /* OP(NORM1) */                                                              \
-  /* OP(NORM2) */                                                              \
-  /* OP(MUL_NO_ZEROS) */
+  OP(AMAX)                                                                     \
+  OP(AVG)                                                                      \
+  OP(NORM1)                                                                    \
+  OP(NORM2)                                                                    \
+  OP(MUL_NO_ZEROS)
 
 class ReductionAttr : public AttributesCRTP<ReductionAttr> {
 public:
--- a/include/fusilli/attributes/reduction_attributes.h
+++ b/include/fusilli/attributes/reduction_attributes.h
@@ -27,7 +27,7 @@ namespace fusilli {
 #define FUSILLI_REDUCTION_MODES(OP)                                            \
   OP(NOT_SET)                                                                  \
   OP(ADD)                                                                      \
-  OP(SUM)                                                                      \
+  OP(SUM) /* Remove once SUM is deprecated  */                                 \
   OP(MUL)                                                                      \
   OP(MIN)                                                                      \
   OP(MAX)                                                                      \
--- a/include/fusilli/support/asm_emitter.h
+++ b/include/fusilli/support/asm_emitter.h
@@ -2156,6 +2156,150 @@ inline std::string ReductionNode::emitNodePreAsm() const {
                                                     torch.aten.sum.dim_IntList)
     FUSILLI_DECLARE_KEEPDIM_REDUCTION_EMITTER(MIN, torch.aten.amin)
     FUSILLI_DECLARE_KEEPDIM_REDUCTION_EMITTER(MAX, torch.aten.amax)
+  case ReductionAttr::Mode::NORM1: {
+    constexpr std::string_view schema = R"(
+    {0}
+    {1}
+    %abs_{2} = torch.aten.abs {4} : {5} -> {5}
+    %keepdim_{2} = torch.constant.bool true
+    %dtype_{2} = torch.constant.none
+    {3}_{2}_perm = torch.aten.sum.dim_IntList %abs_{2}, %reduction_dims_{2}, %keepdim_{2}, %dtype_{2} : {5}, !torch.list<int>, !torch.bool, !torch.none -> {6}
+    {7}
+    )";
+
+    return std::format(schema,
+                       permuteX,             // {0}
+                       dimListOss.str(),     // {1}
+                       suffix,               // {2}
+                       getResultNamesAsm(),  // {3}
+                       getOperandNamesAsm(), // {4}
+                       getOperandTypesAsm(), // {5}
+                       getResultTypesAsm(),  // {6}
+                       permuteY              // {7}
+    );
+  }
+  case ReductionAttr::Mode::AMAX: {
+    constexpr std::string_view schema = R"(
+    {0}
+    {1}
+    %abs_{2} = torch.aten.abs {4} : {5} -> {5}
+    %keepdim_{2} = torch.constant.bool true
+    {3}_{2}_perm = torch.aten.amax %abs_{2}, %reduction_dims_{2}, %keepdim_{2} : {5}, !torch.list<int>, !torch.bool -> {6}
+    {7}
+    )";
+
+    return std::format(schema,
+                       permuteX,             // {0}
+                       dimListOss.str(),     // {1}
+                       suffix,               // {2}
+                       getResultNamesAsm(),  // {3}
+                       getOperandNamesAsm(), // {4}
+                       getOperandTypesAsm(), // {5}
+                       getResultTypesAsm(),  // {6}
+                       permuteY              // {7}
+    );
+  }
+  case ReductionAttr::Mode::AVG: {
+    constexpr std::string_view schema = R"(
+    {0}
+    {1}
+    %keepdim_{2} = torch.constant.bool true
+    %dtype_{2} = torch.constant.none
+    {3}_{2}_perm = torch.aten.mean.dim {4}, %reduction_dims_{2}, %keepdim_{2}, %dtype_{2} : {5}, !torch.list<int>, !torch.bool, !torch.none -> {6}
+    {7}
+    )";
+
+    return std::format(schema,
+                       permuteX,             // {0}
+                       dimListOss.str(),     // {1}
+                       suffix,               // {2}
+                       getResultNamesAsm(),  // {3}
+                       getOperandNamesAsm(), // {4}
+                       getOperandTypesAsm(), // {5}
+                       getResultTypesAsm(),  // {6}
+                       permuteY              // {7}
+    );
+  }
+  case ReductionAttr::Mode::NORM2: {
+    constexpr std::string_view schema = R"(
+    {0}
+    {1}
+    %sq_{2} = torch.aten.mul.Tensor {4}, {4} : {5}, {5} -> {5}
+    %keepdim_{2} = torch.constant.bool true
+    %dtype_{2} = torch.constant.none
+    %sumsq_{2} = torch.aten.sum.dim_IntList %sq_{2}, %reduction_dims_{2}, %keepdim_{2}, %dtype_{2} : {5}, !torch.list<int>, !torch.bool, !torch.none -> {6}
+    {3}_{2}_perm = torch.aten.sqrt %sumsq_{2} : {6} -> {6}
+    {7}
+    )";
+
+    return std::format(schema,
+                       permuteX,             // {0}
+                       dimListOss.str(),     // {1}
+                       suffix,               // {2}
+                       getResultNamesAsm(),  // {3}
+                       getOperandNamesAsm(), // {4}
+                       getOperandTypesAsm(), // {5}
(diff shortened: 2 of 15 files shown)


include/fusilli/attributes/reduction_attributes.h
```diff
@@ -27,15 +27,14 @@ namespace fusilli {
 #define FUSILLI_REDUCTION_MODES(OP)                                            \
   OP(NOT_SET)                                                                  \
   OP(SUM)                                                                      \
-  /* OP(ADD) */                                                                \
```
sjain-stanford, line 30: Is `ADD` dropped because it's equivalent to SUM or some other reason?

include/fusilli/support/asm_emitter.h
```diff
@@ -1911,20 +1961,64 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+    return std::format(kMulNoZerosSchema,
+                       permuteX,             // {0}
+                       dimListOss.str(),     // {1}
+                       suffix,               // {2}
+                       getResultNamesAsm(),  // {3}
+                       getOperandNamesAsm(), // {4}
+                       getOperandTypesAsm(), // {5}
+                       getResultTypesAsm(),  // {6}
+                       permuteY,             // {7}
+                       boolType              // {8}
```
sjain-stanford: hyper nit: Use `/*  {0}  */` style comments in format string placeholders for consistency with the rest.

include/fusilli/support/asm_emitter.h
```diff
@@ -1891,6 +1891,56 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+    %sumsq_{2} = torch.aten.sum.dim_IntList %sq_{2}, %reduction_dims_{2}, %keepdim_{2}, %dtype_{2} : {5}, !torch.list<int>, !torch.bool, !torch.none -> {6}
+    {3}_{2}_perm = torch.aten.sqrt %sumsq_{2} : {6} -> {6}
+    {7}
+    )";
+
+  constexpr std::string_view kMulSchema = R"(
+    {0}
+    {1}
+    %dtype_{2} = torch.constant.none
+    {3}_{2}_perm = torch.prims.prod {4}, %reduction_dims_{2}, %dtype_{2} : {5}, !torch.list<int>, !torch.none -> {6}
```
sjain-stanford, line 2192: Claude tells me this is not registered in the torch dialect. Is that true? 

Do we need to use `torch.aten.prod.dim_int` (with chaining for multiple reduction dims)?

sjain-stanford (changes requested):
Might need rebase / CI seems unclean.

include/fusilli/support/asm_emitter.h
rsuderman: It exists, it was only added recently so its likely unaware it exists today.

include/fusilli/attributes/reduction_attributes.h
rsuderman: Removed because redundancy. Added it back in so we can eventually delete SUM to match the hipdnn naming conventions.

include/fusilli/support/asm_emitter.h
rsuderman: Fixed comment.

include/fusilli/attributes/reduction_attributes.h
```diff
@@ -26,16 +26,16 @@ namespace fusilli {
 
 #define FUSILLI_REDUCTION_MODES(OP)                                            \
   OP(NOT_SET)                                                                  \
+  OP(ADD)                                                                      \
   OP(SUM)                                                                      \
```
sjain-stanford: I'd drop SUM from the op list / samples / lit tests, and add proper coverage for ADD (samples, lit tests) instead to match with cudnn/hipdnn.

include/fusilli/support/asm_emitter.h
```diff
@@ -2175,11 +2225,55 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+                       getOperandTypesAsm(), /* {5} */                         \
+                       getResultTypesAsm(),  /* {6} */                         \
+                       permuteY              /* {7} */                         \
+    );                                                                         \
+  }
+
   switch (reductionAttr.getMode()) {
+  case ReductionAttr::Mode::ADD:
     FUSILLI_DECLARE_KEEPDIM_DTYPE_REDUCTION_EMITTER(SUM,
                                                     torch.aten.sum.dim_IntList)
```
sjain-stanford: It's a little subtle that `ADD` and `SUM` share the same body. We can drop SUM and just instantiate the macro for ADD directly.

include/fusilli/support/asm_emitter.h
```diff
@@ -2155,6 +2155,56 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+  constexpr std::string_view kNorm2Schema = R"(
+    {0}
+    {1}
+    %sq_{2} = torch.aten.mul.Tensor {4}, {4} : {5}, {5} -> {5}
+    %keepdim_{2} = torch.constant.bool true
+    %dtype_{2} = torch.constant.none
+    %sumsq_{2} = torch.aten.sum.dim_IntList %sq_{2}, %reduction_dims_{2}, %keepdim_{2}, %dtype_{2} : {5}, !torch.list<int>, !torch.bool, !torch.none -> {6}
+    {3}_{2}_perm = torch.aten.sqrt %sumsq_{2} : {6} -> {6}
+    {7}
+    )";
```
sjain-stanford: nit: keep kNorm2Schema right after kNorm1Schema for readability

samples/reduction/reduction_ops.cpp
```diff
@@ -48,7 +88,13 @@ TEST_CASE("Reduction ops", "[reduction][graph]") {
...
       ReductionAttr::Mode::SUM,
       ReductionAttr::Mode::MIN,
-      ReductionAttr::Mode::MAX);
+      ReductionAttr::Mode::MAX,
+      ReductionAttr::Mode::NORM1,
+      ReductionAttr::Mode::AMAX,
+      ReductionAttr::Mode::AVG,
+      ReductionAttr::Mode::NORM2,
+      ReductionAttr::Mode::MUL,
+      ReductionAttr::Mode::MUL_NO_ZEROS);
```
sjain-stanford, line 98: Please add `ReductionAttr::Mode::ADD` to the sample generator since it is user-facing now and needs its own coverage.

samples/reduction/reduction_ops.cpp
```diff
@@ -163,13 +247,22 @@ TEST_CASE("Reduction ops", "[reduction][graph]") {
...
       return T(0);
     }
   };
 
-  // int32
-  execute(handle, DataType::Int32, getInitValue.template operator()<int>());
+  // torch.aten.mean.dim is not defined on integer tensors, so AVG only
+  // exercises the floating-point paths.
+  if (mode != ReductionAttr::Mode::AVG)
+    execute(handle, DataType::Int32, getInitValue.template operator()<int>());
```
sjain-stanford, line 268: Don't special case this here. Fusilli should reject integer `AVG` during `ReductionNode` validation, unless there's another way to lower it. Right now it silently accepts the graph until compile time, which is a rough edge.

include/fusilli/support/asm_emitter.h
```diff
@@ -2175,11 +2225,55 @@ inline std::string ReductionNode::emitNodePreAsm() const {
...
+    );                                                                         \
+  }
+
   switch (reductionAttr.getMode()) {
+  case ReductionAttr::Mode::ADD:
     FUSILLI_DECLARE_KEEPDIM_DTYPE_REDUCTION_EMITTER(SUM,
                                                     torch.aten.sum.dim_IntList)
     FUSILLI_DECLARE_KEEPDIM_REDUCTION_EMITTER(MIN, torch.aten.amin)
     FUSILLI_DECLARE_KEEPDIM_REDUCTION_EMITTER(MAX, torch.aten.amax)
+    FUSILLI_DECLARE_KEEPDIM_DTYPE_REDUCTION_EMITTER(AVG, torch.aten.mean.dim)
```
sjain-stanford, line 2247: A user can still build an Int32 AVG graph that passes `graph->validate()` and then fails during compile. I reproduced this locally by compiling an Int32 variant of the emitted AVG IR; the Float variant compiles, while the Int32 variant fails in `iree-compile`.

Preferably reject integral/bool input or output dtypes for AVG in `ReductionNode::postValidateNode()` with an `InvalidAttribute` error, or implement an integer-compatible AVG lowering.

tests/CMakeLists.txt
```diff
@@ -225,7 +225,13 @@ add_fusilli_lit_tests(
     lit/test_sdpa_asm_emitter_cross_attn.cpp
     lit/test_reduction_asm_emitter_sum.cpp
```
sjain-stanford: Let's update + rename this LIT test to be for `add` instead.

samples/reduction/reduction_ops.cpp
```diff
@@ -48,7 +88,13 @@ TEST_CASE("Reduction ops", "[reduction][graph]") {
   const auto mode = GENERATE(
       ReductionAttr::Mode::SUM,
       ReductionAttr::Mode::MIN,
-      ReductionAttr::Mode::MAX);
+      ReductionAttr::Mode::MAX,
+      ReductionAttr::Mode::NORM1,
+      ReductionAttr::Mode::AMAX,
+      ReductionAttr::Mode::AVG,
+      ReductionAttr::Mode::NORM2,
```
sjain-stanford: same nit: NORM1 -> NORM2

samples/reduction/reduction_ops.cpp
sjain-stanford: I guess execute would fail early (at graph validate phase) instead of at iree-compile time so this special case might still be needed. But it's cleaner to have it caught during pre-compile validation.

include/fusilli/attributes/reduction_attributes.h
rsuderman: We should do this in the follow up. The reason I added both was so we could migrate the fusilli plugin to the correct naming, then delete once we have migrated over.

include/fusilli/support/asm_emitter.h
rsuderman: Switched but you should see clang-tidy failing due to consecutive identical switch statements.

include/fusilli/support/asm_emitter.h
rsuderman: Fixed.

samples/reduction/reduction_ops.cpp
rsuderman: Changed to validation level failure. Still have the conditional sample to avoid test failure due to validation.

tests/CMakeLists.txt
rsuderman: Renamed.

include/fusilli/support/asm_emitter.h
rsuderman: Changed to validation level failure.

samples/reduction/reduction_ops.cpp
rsuderman: done.

include/fusilli/attributes/reduction_attributes.h
sjain-stanford: That's reasonable. Can you add a TODO here?

samples/reduction/reduction_ops.cpp
```diff
@@ -157,19 +243,29 @@ TEST_CASE("Reduction ops", "[reduction][graph]") {
...
       return std::numeric_limits<T>::max();
     case ReductionAttr::Mode::MAX:
       return std::numeric_limits<T>::lowest();
+    case ReductionAttr::Mode::MUL:
+    case ReductionAttr::Mode::MUL_NO_ZEROS:
+      return T(1);
+    case ReductionAttr::Mode::AMAX:
+    case ReductionAttr::Mode::AVG:
+    case ReductionAttr::Mode::NORM1:
+    case ReductionAttr::Mode::NORM2:
```
sjain-stanford, line 259: nit: move these four cases above, alongside ADD and SUM (which also `return T(0);`), or move cases for ADD and SUM here if you want it together with the default case.

include/fusilli/support/asm_emitter.h
sjain-stanford: Yep that only works if we were removing SUM, not keeping both. Let's go back to what you had since it'll get cleaned up in a follow-on anyway.

sjain-stanford (approved):
LGTM modulo two more comments.

include/fusilli/attributes/reduction_attributes.h
rsuderman: Added

include/fusilli/support/asm_emitter.h
rsuderman: Done.

samples/reduction/reduction_ops.cpp
rsuderman: Moved.

PR #327, opened 2026-04-09
Support independent K/V head counts in SDPA test utilities

keshavvinayak01:
Split headsKV into headsK/headsV in referenceSdpa and executeSdpa to allow testing GQA with H_k != H_v. headsV defaults to headsK for backward compatibility. Add test for Q=8, K=4, V=2 heads.


keshavvinayak01 (commented):
Putting this as draft for now because first iree needs to bump torch-mlir to include: https://github.com/llvm/torch-mlir/pull/4523, then we'll need to bump IREE, only then the CI will pass :)

sjain-stanford:
While we were waiting, I landed the built-in SDPA op so here's another one that relaxes this constraint for both custom and built-in SDPA ops: https://github.com/iree-org/fusilli/pull/354. Closing this for now.

PR #332, opened 2026-04-10
Add broadcast-compatible scale support for RmsNorm

rsuderman:
Allow RmsNorm scale tensors to have broadcast-compatible shapes (e.g. {1,c,1,1} instead of {1,c,h,w}), letting torch.aten.rms_norm handle the broadcasting internally. Also update getScalarConstantAsm to emit the actual tensor shape instead of hardcoding tensor<1x...>.

### DIFF
--- a/include/fusilli/node/norm_utils.h
+++ b/include/fusilli/node/norm_utils.h
@@ -98,6 +98,34 @@ inline void inferScaleBiasDimAndStride(std::shared_ptr<TensorAttr> &tensor,
   inferStride(tensor, getScaleBiasStride(tensor->getDim(), xStride));
 }
 
+// Returns true if scaleDim is broadcast-compatible with the expected
+// scale/bias shape derived from xDim. Each dimension must either match
+// the corresponding X dimension (with batch forced to 1) or be 1.
+inline bool isScaleBiasBroadcastCompatible(const std::vector<int64_t> &scaleDim,
+                                           const std::vector<int64_t> &xDim,
+                                           size_t batchDim = 0) {
+  if (scaleDim.size() != xDim.size())
+    return false;
+  for (size_t i = 0; i < scaleDim.size(); ++i) {
+    int64_t expected = (i == batchDim) ? 1 : xDim[i];
+    if (scaleDim[i] != expected && scaleDim[i] != 1)
+      return false;
+  }
+  return true;
+}
+
+// Infers dim and stride of a scale tensor for RmsNorm.
+// If dims are not set, they default to X's shape with batch=1.
+// If dims are already set (e.g. broadcast-compatible {1,c,1,1}),
+// they are preserved and strides are computed for that shape.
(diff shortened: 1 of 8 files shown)


sjain-stanford (changes requested):
In isolation this change is fine but I'm not sure this is a level of flexibility we _should_ allow. Both PT and cudnn consistently use per element scale for RMS norm (i.e. of normalized shape - matching input shape modulo batch dims). Why would we want to differ / be more general than that? If this change was prompted by allowing hipdnn -> fusilli to bridge, then its a placebo and the real issue lies in hipdnn scale not matching normalized shapes.

rsuderman:
> In isolation this change is fine but I'm not sure this is a level of flexibility we _should_ allow. Both PT and cudnn consistently use per element scale for RMS norm (i.e. of normalized shape - matching input shape modulo batch dims). Why would we want to differ / be more general than that? If this change was prompted by allowing hipdnn -> fusilli to bridge, then its a placebo and the real issue lies in hipdnn scale not matching normalized shapes.

Docs and samples seem to support both paths. I imagine this is a case of feature creep between the API and docs. From a fusilli standpoint its pretty trivial to support both so we should probably just do that.

sjain-stanford:
> Docs and samples seem to support both paths.

There is no scenario I am possibly seeing where the scale is of reduced shape (and needs broadcasting). It is a learned parameter, initialized to the normalized shape: https://github.com/pytorch/pytorch/blob/70d99e998b4955e0049d13a98d77ae1b14db1f45/torch/nn/modules/normalization.py#L402.

cudnn [sample](https://github.com/NVIDIA/cudnn-frontend/blob/main/samples/cpp/norm/rmsnorm.cpp#L44)'s scale shape is also consistent with the input's normalized shape (don't get tricked by the `{1, hidden_size, 1, 1}` -> it's the same as `X`'s normalized shape `{batch_size * seq_length, hidden_size, 1, 1}`. 
```
    auto X     = graph.tensor(fe::graph::Tensor_attributes()
                              .set_name("X")
                              .set_data_type(fe::DataType_t::FLOAT)
                              .set_dim({batch_size * seq_length, hidden_size, 1, 1})
                              .set_stride({hidden_size, 1, hidden_size, hidden_size}));
    auto scale = graph.tensor(fe::graph::Tensor_attributes()
                                  .set_name("scale")
                                  .set_dim({1, hidden_size, 1, 1})
                                  .set_stride({hidden_size, 1, hidden_size, hidden_size})
                                  .set_data_type(fe::DataType_t::FLOAT));
```

Hence, at the moment I'm not seeing a use-case for supporting broadcasting. Doing so just-in-case is dangerous: it can mask bugs like the one in hipdnn that you're fixing as we speak - these scare me more because they can cause numerical issues if the shapes match up via broadcasting but the numerics of scaling are wrong / inconsistent with the training framework that produced them. So I'm still of the opinion we should NOT support this in fusilli. 

Perhaps we can park this for now and re-visit if we come across a production use-case where this is needed?

PR #337, opened 2026-04-14
Add IDENTITY, RSQRT, SIN, SQRT, and TAN pointwise op support

rsuderman:
Enable five additional unary pointwise operations with ASM emitter support, sample tests, and lit tests.


AaronStGeorge (approved):
LGTM

Issue #255, opened 2026-03-17
tuner <> fusilli V1: Add an option for fusilli to invoke the iree tuner directly


bangtianliu:
V1: add an option for fusilli to invoke the tuner directly so that non-expert end users can use it

Tracking issue [WIP]

kuhar:
cc: @rsuderman

PR #339, opened 2026-04-14
Fusilli <> tuner v1: Add IREE tuner integration via Python wrapper script

bangtianliu:
This PR adds `benchmarks/run_tuner.py`, a Python wrapper around `amdsharktuner.libtuner` that lets non-expert users invoke the IREE tuner directly from Fusilli.

Issue: https://github.com/iree-org/fusilli/issues/255

Assisted-by: [Claude Code](https://claude.ai/code)


sjain-stanford:
I think this is a useful next step but I am pretty confused why you'd want to integrate it at the fusilli core layer when all the direct consumers of this will be at the [fusilli provider](https://github.com/ROCm/rocm-libraries/tree/develop/dnn-providers/fusilli-provider) layer. That's the main user facing interface for fusilli (via hipdnn) and this fits there. @AaronStGeorge could you share what would be a good interface point and next steps to get this integrated properly?

bangtianliu:
> I think this is a useful next step but I am pretty confused why you'd want to integrate it at the fusilli core layer when all the direct consumers of this will be at the [fusilli provider](https://github.com/ROCm/rocm-libraries/tree/develop/dnn-providers/fusilli-provider) layer. That's the main user facing interface for fusilli (via hipdnn) and this fits there. @AaronStGeorge could you share what would be a good interface point and next steps to get this integrated properly?

Thanks for the pointer,  that link is helpful. You're right, the mistake on my end came from a lack of context about where fusilli is actually used.

PR #349, opened 2026-04-16
Change CompileSession and CompileCommand APIs to take backend instead of handle

rsuderman:
Handle requires a runtime to invoke. Compilation should only need the backend target to avoid constructing a full runtime unnecessarily.

### DIFF
--- a/include/fusilli/backend/compile_command.h
+++ b/include/fusilli/backend/compile_command.h
@@ -48,7 +48,7 @@ public:
   // This constructs the full command with:
   // - iree-compile executable path
   // - input file path
-  // - backend-specific flags from Handle
+  // - backend-specific flags
   // - statistics output flags
   // - output file specification
   //
--- a/include/fusilli/backend/compile_command.h
+++ b/include/fusilli/backend/compile_command.h
@@ -14,7 +14,6 @@
 #define FUSILLI_BACKEND_COMPILE_COMMAND_H
 
 #include "fusilli/backend/backend.h"
-#include "fusilli/backend/handle.h"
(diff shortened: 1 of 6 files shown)


include/fusilli/backend/compile_command.h
sjain-stanford: nit: 
```
// - backend-specific flags
```

include/fusilli/backend/compile_session.h
sjain-stanford: Update docstring

```suggestion
//   ErrorOr<CompileSession> session = context->createSession(backend);
```

include/fusilli/backend/handle.h
sjain-stanford: Are these friend declarations still needed? They no longer take a `Handle` parameter and `getBackend()` is now public.

include/fusilli/backend/handle.h
```diff
@@ -119,6 +119,8 @@ class Handle {
   Handle &operator=(Handle &&) noexcept = default;
   ~Handle() = default;
 
+  Backend getBackend() const { return backend_; }
+
   // Allow Graph, Buffer, and CompileCommand objects to access private Handle
   // methods namely `getDevice()`, `getInstance()`, and `getBackend()`.
```
sjain-stanford: Drop `getBackend()`.

sjain-stanford (commented):
Can take another look once addressed.

include/fusilli/backend/handle.h
rsuderman: Done

sjain-stanford (approved):

PR #350, opened 2026-04-17
[Do not merge] Decouple Handle from Graph::compile, introduce explicit load step

sjain-stanford:
This demonstrates syntactic sugar API re-arrangement to remove `handle` dependency from `Graph::compile`. It shifts the VM context load step to its own `Graph::load` API. This doesn't inherently remove the handle dependency from the one-time setup that runs outside the hot execute loop. It just makes it so the `compile` in isolation looks clean with no handle requirement.

I personally think this is NOT a change worth landing. It is 
1) incomplete: Even without the handle dependency, we still don't inherently support compiling for multiple backends and having the execute pick the one that matches
1) unnecessary: Adds an extra user-facing `graph->load` call for no good reason - the current approach hides this underneath the compile call as loading the compiled bytecode into the VM session and setup is an implementation detail that serves no good purpose being exposed to the user

Perhaps at a later point we might support multi-backend compilations and caching - this decoupling of handle from compile is best revisited with the new architecture in place.

### DIFF
--- a/include/fusilli/graph/graph.h
+++ b/include/fusilli/graph/graph.h
@@ -111,12 +111,13 @@ public:
     return ok();
   }
 
-  // Compiles the graph using IREE compiler and sets up the IREE VM
-  // context for future g->execute calls.
+  // Compiles the graph using IREE compiler to produce a VMFB artifact.
+  // Does not create any runtime state; call `load(handle)` afterwards to
+  // set up the IREE VM context for execution.
   //
   // Set `remove = true` to remove compilation artifacts (cache files) when
   // this `Graph` instance goes out of scope.
-  ErrorObject compile(const Handle &handle, bool remove = false) {
+  ErrorObject compile(Backend backend, bool remove = false) {
     FUSILLI_LOG_LABEL_ENDL("INFO: Compiling Graph");
     FUSILLI_RETURN_ERROR_IF(!isValidated_, ErrorCode::NotValidated,
                             "Graph must be validated before being compiled");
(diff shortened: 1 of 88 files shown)


include/fusilli/graph/graph.h
```diff
@@ -125,14 +126,33 @@ class Graph : public INode {
...
-                             getCompiledArtifact(handle, generatedAsm, remove));
+    FUSILLI_ASSIGN_OR_RETURN(
+        auto vmfbPath, getCompiledArtifact(backend, generatedAsm, remove));
 
     FUSILLI_LOG_LABEL_ENDL("INFO: Compiled Graph cached at \"" +
                            vmfbPath.string() + "\"");
 
-    // Create per-graph IREE VM context and load the compiled artifact.
-    FUSILLI_CHECK_ERROR(createVmContext(handle, vmfbPath.string()));
+    compiledBackend_ = backend;
```
sjain-stanford, line 135: **Major issue**

- `include/fusilli/graph/graph.h:135`: `compile()` records the new backend and VMFB path but leaves any previously loaded runtime state intact. If a caller does `compile(); load(); compile(); execute();`, `execute()` will still pass the checks in `include/fusilli/backend/runtime.h:374` and `include/fusilli/backend/runtime.h:377` because `vmContext_` and `vmFunction_` still refer to the old loaded artifact. `workspaceSize_` can also remain from the old module, so workspace allocation after the second compile may use stale requirements. This regresses the old `compile(handle)` behavior, where every successful compile created a fresh VM context. Clear the loaded state after a successful new artifact selection, for example reset `vmContext_`, `vmFunction_`, `workspaceSize_`, and `vmInputListCapacity_` when `compile()` updates `compiledBackend_`/`vmfbPath_`, or track a loaded generation and have `execute()` reject stale loads. Add a regression test that compiles and loads a graph, compiles it again, verifies `getWorkspaceSize()` returns `std::nullopt` or otherwise reports not-loaded state, and verifies `execute()` fails until `load()` is called again.

sjain-stanford (commented):
## # PR Review: [Do not merge] Decouple Handle from Graph::compile, introduce explicit load step

Posted 1 inline review comment(s).

### Summary
This PR splits `Graph::compile(handle)` into `Graph::compile(backend)` for VMFB generation and `Graph::load(handle)` for VM context creation and workspace-size discovery. It updates samples, tests, and benchmarks to call `load()` after compilation, and adds basic tests for calling `load()` too early and for backend mismatch.

### Recommendations
Consider keeping a short compatibility path or migration note for downstream users of `Graph::compile(handle)`, since this PR changes the public C++ Graph API across every sample and test. The correctness fix above should come first because it affects runtime behavior even when callers use the new API.

### General / Unplaced Findings

- **Minor issue** (include/fusilli/graph/graph.h:343)
  - `include/fusilli/graph/graph.h:343`: The public `getWorkspaceSize()` comment still says `std::nullopt` means "not compiled", and `include/fusilli/backend/runtime.h:375` / `include/fusilli/backend/runtime.h:378` still tell users `execute()` needs a successful `compile()` first. With the new lifecycle, compile can succeed while the graph is still not loaded, so these diagnostics point users at the wrong fix. Update the comments, test names around `tests/test_graph.cpp:504`, and execute error messages to refer to `load()` or "compiled and loaded" state.

### Verdict
REQUEST_CHANGES

_Review file: `pr-review-fusilli-350.md`_

sjain-stanford:
Superseded by a better change: https://github.com/iree-org/fusilli/pull/377

PR #355, opened 2026-04-17
Add SOFTPLUS_FWD and SWISH_FWD pointwise ops

rsuderman:
Emit torch.aten.softplus with configurable beta and threshold attributes (defaulting to 1.0 and 20.0) and torch.aten.silu for swish. Adds sample and lit tests for both ops.

### DIFF
--- a/samples/pointwise/pointwise_unary_ops.cpp
+++ b/samples/pointwise/pointwise_unary_ops.cpp
@@ -75,7 +75,9 @@ TEST_CASE("Pointwise unary ops", "[pointwise][graph]") {
     case PointwiseAttr::Mode::RSQRT:
     case PointwiseAttr::Mode::SIGMOID_FWD:
     case PointwiseAttr::Mode::SIN:
+    case PointwiseAttr::Mode::SOFTPLUS_FWD:
     case PointwiseAttr::Mode::SQRT:
+    case PointwiseAttr::Mode::SWISH_FWD:
     case PointwiseAttr::Mode::TAN:
     case PointwiseAttr::Mode::TANH_FWD:
       return true;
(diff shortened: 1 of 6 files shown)


keshavvinayak01 (approved):
LGTM

sjain-stanford (commented):
This breaks compat with cudnn / hipdnn's swish for `beta != 1`.  The swish function is not a SILU when `beta !=1`.

Swish: https://en.wikipedia.org/wiki/Swish_function
https://docs.pytorch.org/docs/stable/generated/torch.nn.SiLU.html

Requesting changes (in a follow-on since this landed before I could review):
Include `setSwishBeta` and `getSwishBeta` methods on the attribute (with default 1.0f). Then either 1) reject non-default beta in pre-validate, or 2) emit `aten.sigmoid` + `aten.mul` for `beta != 1`, and `aten.silu` otherwise.

In general, please wait for a comment from either me or one of the other contributors who is familiar with adding new ops (like @IanWood1).

samples/pointwise/pointwise_unary_ops.cpp
```diff
@@ -248,11 +252,25 @@ TEST_CASE("Pointwise unary ops", "[pointwise][graph]") {
       y = std::sin(xD);
       break;
     }
+    case PointwiseAttr::Mode::SOFTPLUS_FWD: {
+      double xD = static_cast<double>(x);
+      // SOFTPLUS(x) = log(1 + exp(x)); for x > threshold, result ~= x.
+      // Matches torch.aten.softplus with beta=1.0 and threshold=20.0.
+      constexpr double threshold = 20.0;
```
sjain-stanford, line 259: nit: use `k`-prefix convention  for constants.. `kSoftplusThreshold`

IanWood1: The k prefix only applies to global constants. We could change the clang-tidy rules to apply to local ones, too. There are some other cases without the prefix. e.g. https://github.com/iree-org/fusilli/blob/aeed46fe35604c9f055f35005b95dc7e8f56633e/include/fusilli/node/conv_node.h#L341

PR #358, opened 2026-04-20
Add ADD_SQUARE pointwise op

rsuderman:
Emits y = x0 + x1^2 as torch.aten.mul.Tensor followed by torch.aten.add.Tensor. Adds sample and lit tests.

### DIFF
--- a/include/fusilli/support/asm_emitter.h
+++ b/include/fusilli/support/asm_emitter.h
@@ -1778,6 +1778,15 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
     {6}
 )";
 
+  constexpr std::string_view kAddSquareSchema = R"(
+    {0}
+    {1}
+    %add_square_sq_{9} = torch.aten.mul.Tensor {3}, {3} : {5}, {5} -> {5}
+    %alpha_{9} = torch.constant.int 1
+    {6} = torch.aten.add.Tensor {2}, %add_square_sq_{9}, %alpha_{9} : {4}, {5}, !torch.int -> {7}
+    {8}
+)";
+
   constexpr std::string_view kIdentitySchema = R"(
     {0}
     %none_{7} = torch.constant.none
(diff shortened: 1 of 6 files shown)


keshavvinayak01 (commented):
Looks fine, I'd wait for @sjain-stanford's review.

samples/pointwise/pointwise_binary_ops.cpp
```diff
@@ -120,6 +121,10 @@ TEST_CASE("Pointwise binary ops", "[pointwise][graph]") {
       y = x0 + x1;
       break;
     }
+    case PointwiseAttr::Mode::ADD_SQUARE: {
+      y = x0 + x1 * x1;
```
sjain-stanford, line 125: Is it `x0 + x1^2` or `(x0 + x1)^2`? This implements the former but [this](https://sourcegraph.com/r/github.com/ROCm/rocm-libraries/-/blob/projects/hipdnn/backend/include/HipdnnPointwiseMode.h?L29) comment in hipdnn seems to indicate latter? Can't tell what cudnn is doing from a cursory look.

include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1903,29 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
...
+  case PointwiseAttr::Mode::ADD_SQUARE: {
+    const auto &in0 = pointwiseAttr.getIN_0();
+    const auto &in1 = pointwiseAttr.getIN_1();
+    const std::string suffix = getName();
+    const std::string in0Name = in0->getValueNameAsm() + "_" + suffix + "_perm";
+    const std::string in1Name = in1->getValueNameAsm() + "_" + suffix + "_perm";
+    const std::string in0Type =
+        in0->getTensorTypeAsm(/*isValueTensor=*/true, /*useLogicalDims=*/true);
+    const std::string in1Type =
+        in1->getTensorTypeAsm(/*isValueTensor=*/true, /*useLogicalDims=*/true);
```
sjain-stanford: Is there something inherently different here that makes using existing utilities (`getOperandNamesAsm` and `getOperandTypesAsm`) not feasible? I'd reuse that if possible.

sjain-stanford (commented):
I have two comments.

samples/pointwise/pointwise_binary_ops.cpp
rsuderman: Hipdnn seems to have missed the goal
https://docs.nvidia.com/deeplearning/cudnn/archives/cudnn-892/api/index.html#:~:text=CUDNN_POINTWISE_ADD_SQUARE,tensor%20is%20computed.

include/fusilli/support/asm_emitter.h
rsuderman: They have to be separated as the operation needs to be decomposed to a mul and add. Moved the declarations to helpers intead.

samples/pointwise/pointwise_binary_ops.cpp
sjain-stanford: Yeah this is a bug in hipdnn (cc: @BrianHarrisonAMD).

sjain-stanford (approved):

samples/pointwise/pointwise_binary_ops.cpp
BrianHarrisonAMD: Ill log an issue, and investigate if ref impl is off as well.

ty ty!

BrianHarrisonAMD: No ref impls yet actually, so just doc fix.

https://github.com/ROCm/rocm-libraries/issues/6753

PR #359, opened 2026-04-20
Add GEN_INDEX pointwise op

rsuderman:
Emits the output as a linalg.generic that materializes the iteration index along a configurable axis (via PointwiseAttr::setGenIdxAxis). The generic is bridged back to torch via torch_c.from_builtin_tensor so the result feeds the standard output permute.

Also adds buildBuiltinTensorTypeStr / getSignlessElementTypeAsm / getIdentityAffineMapAsm / getIteratorTypesAsm helpers and relaxes the lit-test MLIR indentation check to allow nested region bodies.


include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1947,106 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
     FUSILLI_DECLARE_SUB_ADD_TORCH_EMITTER(ADD, torch.aten.add.Tensor)
     FUSILLI_DECLARE_SUB_ADD_TORCH_EMITTER(SUB, torch.aten.sub.Tensor)
 
+<<<<<<< Updated upstream
+=======
```
sjain-stanford: improper merge?

rsuderman: Looks like claude failed on the merge. Fixed.

include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1947,80 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
     FUSILLI_DECLARE_SUB_ADD_TORCH_EMITTER(ADD, torch.aten.add.Tensor)
     FUSILLI_DECLARE_SUB_ADD_TORCH_EMITTER(SUB, torch.aten.sub.Tensor)
 
+  case PointwiseAttr::Mode::GEN_INDEX: {
+    const auto &out0 = pointwiseAttr.getOUT_0();
+    const std::string suffix = getName();
+    const std::vector<int64_t> &outDims = out0->getDim();
+    const int64_t axis = pointwiseAttr.getGenIdxAxis();
+    assert(axis >= 0 && axis < static_cast<int64_t>(outDims.size()) &&
+           "GEN_INDEX axis out of range");
```
IanWood1, line 1948: This assert is OK but we should be checking this in the verifier, too.

include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1947,80 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
...
+    constexpr std::string_view kGenIndexIntSchema = R"(
+    %gen_index_empty_{0} = tensor.empty() : {1}
+    %gen_index_linalg_{0} = linalg.generic {{indexing_maps = [{2}], iterator_types = [{3}]}} outs(%gen_index_empty_{0} : {1}) {{
+    ^bb0(%gen_index_out_{0}: {4}):
+      %gen_index_idx_{0} = linalg.index {5} : index
+      %gen_index_val_{0} = arith.index_cast %gen_index_idx_{0} : index to {4}
+      linalg.yield %gen_index_val_{0} : {4}
+    }} -> {1}
+    {6} = torch_c.from_builtin_tensor %gen_index_linalg_{0} : {1} -> {7}
+    {8})";
```
IanWood1, line 1978: We probably want a test for the int version to make sure the emitted MLIR is correct.

tests/pointwise_utils.h
```diff
@@ -99,6 +99,46 @@ inline ErrorObject testBinaryPointwiseAsmEmitter(const std::string &graphName,
   return ok();
 }
```
IanWood1, line 100: I don't have a good place to put this but can we test this op in pointwise_unary_ops.cpp?

include/fusilli/attributes/pointwise_attributes.h
```diff
@@ -165,6 +172,7 @@ inline const std::unordered_map<PointwiseAttr::Mode, int>
         {PointwiseAttr::Mode::FLOOR, 1},
         {PointwiseAttr::Mode::GELU_APPROX_TANH_FWD, 1},
         {PointwiseAttr::Mode::GELU_FWD, 1},
+        {PointwiseAttr::Mode::GEN_INDEX, 1},
```
IanWood1, line 175: It's a bit strange that it takes an input that it never uses. Is this just for shape inference?

include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1947,80 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
...
+    assert(axis >= 0 && axis < static_cast<int64_t>(outDims.size()) &&
+           "GEN_INDEX axis out of range");
+
+    const size_t rank = outDims.size();
+    const std::string signlessDtype =
+        getSignlessElementTypeAsm(out0->getDataType());
+    const bool isFloat = signlessDtype == "f16" || signlessDtype == "bf16" ||
+                         signlessDtype == "f32" || signlessDtype == "f64";
+    assert((isFloat || signlessDtype.substr(0, 1) == "i") &&
+           "GEN_INDEX only supports integer or float output dtypes");
```
IanWood1: Same, this assert is OK but we should be checking this in the verifier, too.

include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1947,80 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
...
+    const std::vector<int64_t> &outDims = out0->getDim();
+    const int64_t axis = pointwiseAttr.getGenIdxAxis();
+    assert(axis >= 0 && axis < static_cast<int64_t>(outDims.size()) &&
+           "GEN_INDEX axis out of range");
+
+    const size_t rank = outDims.size();
+    const std::string signlessDtype =
+        getSignlessElementTypeAsm(out0->getDataType());
+    const bool isFloat = signlessDtype == "f16" || signlessDtype == "bf16" ||
+                         signlessDtype == "f32" || signlessDtype == "f64";
```
IanWood1: This doesn't cover all float types. Also, this should probably be moved into a helper near the definition of `DataType` so updating is easier.

include/fusilli/support/asm_emitter.h
```diff
@@ -1894,6 +1947,80 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
...
+    assert((isFloat || signlessDtype.substr(0, 1) == "i") &&
```
IanWood1: `signlessDtype.substr(0, 1) == "i"` seems bug prone

include/fusilli/support/asm_emitter.h
```diff
@@ -128,6 +128,59 @@ inline std::string buildTensorTypeStr(const std::vector<int64_t> &dims,
...
+//   rank=3 → "affine_map<(d0, d1, d2) -> (d0, d1, d2)>"
+// Used for linalg.generic indexing_maps where every operand/result accesses
+// the full iteration space in natural order.
+inline std::string getIdentityAffineMapAsm(size_t rank) {
+  std::ostringstream dims;
+  for (size_t i = 0; i < rank; ++i) {
+    if (i > 0)
+      dims << ", ";
+    dims << "d" << i;
+  }
```
IanWood1: Use `interleave()`

include/fusilli/support/asm_emitter.h
```diff
@@ -128,6 +128,59 @@ inline std::string buildTensorTypeStr(const std::vector<int64_t> &dims,
...
+// the given kind (e.g. "parallel" or "reduction"), e.g.
+//   rank=3, kind="parallel" → "\"parallel\", \"parallel\", \"parallel\""
+// Paired with an indexing_maps attribute on a linalg.generic.
+inline std::string getIteratorTypesAsm(size_t rank, std::string_view kind) {
+  std::ostringstream oss;
+  for (size_t i = 0; i < rank; ++i) {
+    if (i > 0)
+      oss << ", ";
+    oss << "\"" << kind << "\"";
+  }
```
IanWood1: Use `interleave()`

include/fusilli/support/asm_emitter.h
```diff
@@ -128,6 +128,59 @@ inline std::string buildTensorTypeStr(const std::vector<int64_t> &dims,
...
+// leading "s"/"u" that kDataTypeToMlirTypeAsm carries on integer types
+// ("si32" → "i32", "ui8" → "i8"). Builtin MLIR tensors and arith ops use
+// signless integers, unlike torch vtensors which preserve signedness.
+inline std::string getSignlessElementTypeAsm(DataType dtype) {
+  std::string elemType = kDataTypeToMlirTypeAsm.at(dtype);
+  if (elemType.size() >= 2 &&
+      (elemType.substr(0, 2) == "si" || elemType.substr(0, 2) == "ui"))
+    elemType = "i" + elemType.substr(2);
+  return elemType;
+}
```
IanWood1: This might be better suited for types.h

include/fusilli/support/asm_emitter.h
```diff
@@ -128,6 +128,59 @@ inline std::string buildTensorTypeStr(const std::vector<int64_t> &dims,
   return oss.str();
 }
 
+// Returns the signless MLIR element type for a DataType, stripping the
```
IanWood1: These new MLIR helpers could use unit tests. They are fairly simple so they should only need a few.

include/fusilli/attributes/pointwise_attributes.h
rsuderman: Correct — IN_0 is only used for shape inference. The Graph::pointwise single-operand overload (graph.h:937) feeds the input's shape through inferPropertiesNode so the OUT_0 dims/strides can be derived without having to set them explicitly. This matches cuDNN's Graph API shape for GEN_INDEX (it takes an input tensor whose shape drives the output shape). The emitted linalg.generic doesn't consume IN_0's values.

include/fusilli/support/asm_emitter.h
rsuderman: Added to the verifier.

include/fusilli/support/asm_emitter.h
rsuderman: Added

tests/pointwise_utils.h
rsuderman: Test added

include/fusilli/support/asm_emitter.h
rsuderman: Added to the verifier.

include/fusilli/support/asm_emitter.h
rsuderman: Switched how we identify the type information to part of the type declaration system.

include/fusilli/support/asm_emitter.h
rsuderman: Updated.

include/fusilli/support/asm_emitter.h
rsuderman: Updated.

include/fusilli/support/asm_emitter.h
rsuderman: Moved along with the other helper functions.

include/fusilli/support/asm_emitter.h
rsuderman: Tests added to a new catch2 suite.

sjain-stanford (commented):
I gave it a cursory look but would like to do another closer review later today or Thursday.

PR #362, opened 2026-04-21
Add swish_beta parameter to SWISH_FWD

rsuderman:
Expands the SWISH_FWD emission from torch.aten.silu (beta=1 only) to the composite x * sigmoid(beta * x) form, driven by a new PointwiseAttr::setSwishBeta (default 1.0). Sample and lit tests set a non-default beta so the new knob is exercised.

### DIFF
--- a/tests/pointwise_utils.h
+++ b/tests/pointwise_utils.h
@@ -99,6 +99,46 @@ inline ErrorObject testBinaryPointwiseAsmEmitter(const std::string &graphName,
   return ok();
 }
 
+inline ErrorObject testSwishPointwiseAsmEmitter(const std::string &graphName,
+                                                const std::string &opName,
+                                                const std::string &mode,
+                                                std::vector<int64_t> inDims,
+                                                float beta) {
+
+  auto graph = std::make_shared<Graph>();
+  graph->setName(graphName);
+  graph->setIODataType(DataType::Float).setComputeDataType(DataType::Float);
+
+  auto xT = createTestTensor("arg0", inDims, graph.get());
+
+  auto pointwiseAttr = PointwiseAttr()
+                           .setMode(PointwiseAttr::Mode::SWISH_FWD)
+                           .setName(opName)
+                           .setSwishBeta(beta);
+
+  auto yT = graph->pointwise(xT, pointwiseAttr);
+
+  yT->setName("result").setOutput(true);
+
+  FUSILLI_CHECK_ERROR(graph->validate());
+
+  if (mode == "default") {
+    FUSILLI_ASSIGN_OR_RETURN(auto generatedAsm, graph->emitAsm());
+    FUSILLI_CHECK_ERROR(checkMlirIndentation(generatedAsm));
+    std::cout << generatedAsm << std::endl;
+  }
+
+  if (mode == "stats") {
+    FUSILLI_ASSIGN_OR_RETURN(Handle handle, Handle::create(kDefaultBackend));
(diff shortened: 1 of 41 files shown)


tests/pointwise_utils.h
```diff
@@ -99,6 +99,46 @@ inline ErrorObject testBinaryPointwiseAsmEmitter(const std::string &graphName,
   return ok();
 }
 
+inline ErrorObject testSwishPointwiseAsmEmitter(const std::string &graphName,
+                                                const std::string &opName,
+                                                const std::string &mode,
+                                                std::vector<int64_t> inDims,
+                                                float beta) {
```
sjain-stanford: Instead of creating a separate utility for swish (and this will keep growing as we add new ops that take custom attributes), consider refactoring `testUnaryPointwiseAsmEmitter` / `testBinaryPointwiseAsmEmitter` to take the pointwise attribute object passed by the caller. Then you may drop `opName` and `pwMode` from these utils in favor of the attr and reuse the same helpers.
```cpp
inline ErrorObject testUnaryPointwiseAsmEmitter(const std::string &graphName,
                                                  const std::string &mode,
                                                  const PointwiseAttr &pointwiseAttr,
                                                  std::vector<int64_t> inDims);
```

sjain-stanford (changes requested):

tests/pointwise_utils.h
rsuderman: Updated to make the interface change. Does have a bunch of one line changes but should be fine.

sjain-stanford (approved):

tests/pointwise_utils.h
```diff
@@ -24,9 +24,8 @@
 namespace fusilli {
 
 inline ErrorObject testUnaryPointwiseAsmEmitter(const std::string &graphName,
-                                                const std::string &opName,
                                                 const std::string &mode,
-                                                PointwiseAttr::Mode pwMode,
+                                                PointwiseAttr &pointwiseAttr,
                                                 std::vector<int64_t> inDims) {
```
keshavvinayak01, line 30: Need to fix another callsite:
```
tests/lit/test_pointwise_asm_emitter_add_square.cpp:60:17: error: no matching function for call to 'testBinaryPointwiseAsmEmitter'
   60 |   auto status = testBinaryPointwiseAsmEmitter(
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/workspace/tests/pointwise_utils.h:60:20: note: candidate function not viable: requires 5 arguments, but 6 were provided
   60 | inline ErrorObject testBinaryPointwiseAsmEmitter(const std::string &graphName,
      |                    ^                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   61 |                                                  const std::string &mode,
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~
   62 |                                                  PointwiseAttr &pointwiseAttr,
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   63 |                                                  std::vector<int64_t> lhsDims,
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   64 |                                                  std::vector<int64_t> rhsDims) {
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

keshavvinayak01 (approved):

tests/pointwise_utils.h
sjain-stanford: My merge from main broke this.

rsuderman: Fix is landed below.

PR #365, opened 2026-04-22
Add BINARY_SELECT pointwise op

rsuderman:
Emit torch.aten.where.self for a three-input pointwise select that picks between IN_1 and IN_2 based on a boolean IN_0 predicate. Adds a ternary Graph::pointwise overload, a ternary torch emitter schema, and sample and lit tests.

### DIFF
--- a/samples/pointwise/pointwise_ternary_ops.cpp
+++ b/samples/pointwise/pointwise_ternary_ops.cpp
@@ -1,0 +1,153 @@
+// Copyright 2026 Advanced Micro Devices, Inc.
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <fusilli.h>
+
+#include "utils.h"
+
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+
+#include <cstddef>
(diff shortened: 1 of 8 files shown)


samples/pointwise/pointwise_ternary_ops.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+TEST_CASE("Pointwise ternary ops", "[pointwise][graph]") {
+  const auto dims = std::vector<std::vector<int64_t>>{
+      std::vector<int64_t>{2, 16, 64, 64}, std::vector<int64_t>{2, 16, 64, 64},
+      GENERATE(std::vector<int64_t>{2, 16, 64, 64},
+               std::vector<int64_t>{1, 16, 1, 1})};
+
+  // clang-format off
+  const auto mode = GENERATE(PointwiseAttr::Mode::BINARY_SELECT);
+  // clang-format on
```
sjain-stanford, line 48: nit: remove clang-format exclusions (can be added later if/when more ternary ops are added)

sjain-stanford (approved):

samples/pointwise/pointwise_ternary_ops.cpp
rsuderman: Will follow up and fix. It was on auto-merge.

rsuderman: https://github.com/iree-org/fusilli/pull/375

PR #367, opened 2026-04-23
Add GELU_BWD and GELU_APPROX_TANH_BWD pointwise ops

rsuderman:
Emit torch.aten.gelu_backward with configurable approximate mode ("none" for GELU_BWD, "tanh" for GELU_APPROX_TANH_BWD). Adds a binary backward schema that threads grad_output and self through the existing pointwise pipeline, plus sample and lit tests for both ops.


sjain-stanford:
Might need a rebase and fixes for CI.

include/fusilli/support/asm_emitter.h
```diff
@@ -1862,6 +1862,14 @@ inline std::string PointwiseNode::emitNodePreAsm() const {
     {5}
 )";
 
+  constexpr std::string_view kGeluBwdSchema = R"(
+    {0}
+    {1}
+    %gelu_approximate_{8} = torch.constant.str "{9}"
+    {2} = {7} {3}, %gelu_approximate_{8} : {4}, !torch.str -> {5}
```
sjain-stanford: Let's inline the op name (`torch.aten.gelu_backward`) here instead of in the format string below, since it is fixed. This is also consistent with how other schemas are defined.

sjain-stanford (approved):
LGTM modulo one change.

include/fusilli/support/asm_emitter.h
rsuderman: Done.

PR #374, opened 2026-04-27
Add RELU_BWD pointwise op

rsuderman:
Emit torch.aten.threshold_backward with threshold=0 taking (grad_output, self). Adds emitter case, sample test, and minimal lit test.

### DIFF
--- a/samples/pointwise/pointwise_relu_bwd.cpp
+++ b/samples/pointwise/pointwise_relu_bwd.cpp
@@ -24,8 +24,8 @@ using namespace fusilli;
 // Based on parameters, generates a unique name for the graph
 static std::string generateName(DataType type,
                                 const std::vector<int64_t> &dim) {
-  std::string name = std::format("pointwise_relu_bwd_dt{}",
-                                 kDataTypeToMlirTypeAsm.at(type));
+  std::string name =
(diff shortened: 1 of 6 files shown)


samples/pointwise/pointwise_relu_bwd.cpp
```diff
@@ -0,0 +1,125 @@
...
+      REQUIRE(static_cast<float>(val) == static_cast<float>(expected));
+  };
+
+  // Create handle for the target backend.
+  FUSILLI_REQUIRE_ASSIGN(Handle handle, Handle::create(kDefaultBackend));
+
+  // Positive x case: dx = dy.
+  execute(handle, half(2.5f), half(1.25f));
+  // Negative x case: dx = 0.
+  execute(handle, half(2.5f), half(-1.25f));
```
sjain-stanford, line 124: nit: consider adding a zero-input case to explicitly pin the threshold boundary behavior (`x == 0` should produce zero gradient).

sjain-stanford (approved):
LGTM modulo a zero-input sample.

samples/pointwise/pointwise_relu_bwd.cpp
rsuderman: Added

PR #377, opened 2026-04-27
Add explicit AOT compile/load APIs

### LINKED ISSUE #346: Fusilli compile API currently requires a runtime handle
## Intent

Fusilli graph compilation should not require a runtime HIP stream. A compile
operation needs target backend/device information, but it should not force the
caller to create the same runtime handle that will later own execution stream
ordering.

This matters for benchmarking and integration code that wants to keep the
compile phase separate from the timed execution phase:

- Build and validate a graph.
- Compile or retrieve a cached VMFB for the target backend.
- Later create or attach to a runtime stream.
- Execute and time only runtime work on that stream.

Today the public API shape in this checkout is:

```c++
ErrorObject Graph::compile(const Handle &handle, bool remove = false);
```

and the implementation uses that handle for two distinct concerns:

- compile command/session setup through `CompileSession::build(handle, ...)`,
  which only needs backend-derived target flags;
- VM context creation through `createVmContext(handle, vmfbPath)`, which binds
  the compiled artifact to an IREE HAL device created by the handle.

The second concern means callers cannot compile without first creating a
runtime-capable handle. For the AMDGPU backend, creating the runtime handle may
also mean choosing a HIP stream before compilation.

## Relevant Files

- `fusilli/include/fusilli/graph/graph.h`
  - `Graph::compile(const Handle &handle, bool remove = false)`
  - `Graph::getCompiledArtifact(const Handle &handle, ...)`
  - default compile path uses `CompileSession::build(handle, ...)`
- `fusilli/include/fusilli/backend/compile_session.h`
  - `CompileContext::createSession(const Handle &handle)`
  - `CompileSession::build(const Handle &handle, ...)`
- `/home/mahesh/iree/fusilli/include/fusilli/backend/runtime.h`
  - `Graph::createVmContext(const Handle &handle, const std::string &vmfbPath)`

## Minimal Reproducer Shape

This is the shape that integration code would like to express:

```c++
auto graph = std::make_shared<fusilli::Graph>();
graph->setName("compile_without_runtime_stream")
    .setIODataType(fusilli::DataType::BFloat16)
    .setComputeDataType(fusilli::DataType::Float);

auto a = graph->tensor(fusilli::TensorAttr()
                           .setName("a")
                           .setDim({1024, 1024})
                           .setStride({1024, 1}));
auto b = graph->tensor(fusilli::TensorAttr()
                           .setName("b")
                           .setDim({1024, 1024})
                           .setStride({1024, 1}));

auto c = graph->matmul(a, b, fusilli::MatmulAttr().setName("matmul"));
c->setOutput(true);

FUSILLI_CHECK_ERROR(graph->validate());

// Desired: compile for AMDGPU without choosing a HIP execution stream.
// This could be one of:
//   graph->compile(fusilli::Backend::AMDGPU);
//   graph->compile(fusilli::CompileOptions{.backend = fusilli::Backend::AMDGPU});
//   graph->getCompiledArtifact(fusilli::Backend::AMDGPU, asmText, ...);
//
// Current API requires:
FUSILLI_ASSIGN_OR_RETURN(
    auto handle,
    fusilli::Handle::create(fusilli::Backend::AMDGPU,
                            /*deviceId=*/0,
                            reinterpret_cast<std::uintptr_t>(stream)));
FUSILLI_CHECK_ERROR(graph->compile(handle));
```

## Current Workaround

Currently user has to create the HIP stream and AMDGPU runtime handle
before calling `Graph::compile(handle)`, then reuse that same handle for buffer
allocation and execution. This works, but it couples compile setup to stream
selection:

```c++
hipStream_t stream = nullptr;
HIP_CHECK(hipStreamCreate(&stream));
fusilli::Handle handle = createRuntimeHandle(stream);

FUSILLI_CHECK_ERROR(graph->validate());
FUSILLI_CHECK_ERROR(graph->compile(handle));
```

## Requested Direction

Please consider splitting compile target configuration from runtime context
creation. Possible API directions:

- Add a compile-target object that supplies backend/device/compiler flags
  without owning a HIP stream.
- Let `getCompiledArtifact`/compile cache generation take backend or compile
  options, while VM context creation happens separately when a runtime handle is
  supplied.

The goal is not to avoid choosing a backend or device target. The goal is to
avoid requiring a HIP execution stream during compiler setup.

sjain-stanford:
Adds explicit AOT compile/load APIs so callers can compile backend-specific VMFBs separately from runtime loading while keeping `Graph::compile(handle)` as the JIT convenience path over `compileToArtifact(handle.getBackend())` + `loadFromArtifact(handle, vmfbBytes)`.

Summary:
- Updates `Graph::compileToArtifact(Backend)` to return caller-owned raw VMFB bytes. Internally it still uses the compile-side artifact path from `getCompiledArtifact()` and reads the VMFB before returning.
- Updates `Graph::loadFromArtifact(Handle, bytes)` to load from an in-memory VMFB buffer instead of a filesystem path.
- Keeps loaded VMFB bytes alive for the lifetime of the IREE VM context and explicitly tears down runtime state before clearing those bytes.
- Updates the AOT single-backend and multi-backend samples to pass artifacts between compile and execute phases with in-memory `std::vector<uint8_t>` buffers instead of temporary filesystem paths.
- Documents AOT constraints: VMFBs are backend-specific, compile-side cache ownership is not a persistent artifact contract, and callers should keep or serialize byte buffers they need to use later.
- Adds focused graph tests covering compile-only artifacts, load/execute round trips, runtime-state replacement, artifact lifetime across graph instances, and backend mismatch errors.

### DIFF
--- a/include/fusilli/graph/graph.h
+++ b/include/fusilli/graph/graph.h
@@ -174,7 +174,8 @@ public:
   //
   // A `Graph` owns one loaded runtime state at a time. Loading a new artifact
   // replaces any previously loaded VM context, function, workspace size, and VM
-  // input list capacity. To keep multiple backends loaded concurrently, use one
+  // input list capacity. If loading fails, the `Graph` is left without loaded
+  // runtime state. To keep multiple backends loaded concurrently, use one
   // `Graph` instance per backend artifact. To switch backends on one `Graph`,
   // call `loadFromArtifact()` with the selected backend artifact before
   // `execute()`.
@@ -304,8 +305,7 @@ public:
   Graph &operator=(const Graph &) = delete;
   Graph(Graph &&) noexcept = default;
   Graph &operator=(Graph &&) noexcept = default;
-  // The VM context may retain references into loadedArtifactBytes_.
-  ~Graph() { clearRuntimeState(); }
+  ~Graph() = default;
 
   // Getters and setters for graph context.
   const std::string &getName() const override final {
@@ -465,11 +465,8 @@ private:
   ErrorObject createVmContext(const Handle &handle);
 
   void clearRuntimeState() {
-    // Teardown order matters: the IREE bytecode module owned by vmContext_ may
-    // retain references into loadedArtifactBytes_, so release vmContext_ before
-    // clearing the backing VMFB bytes.
-    vmContext_.reset();
     vmFunction_.reset();
+    vmContext_.reset();
     workspaceSize_.reset();
     loadedBackend_.reset();
     loadedArtifactBytes_.clear();
@@ -694,10 +691,14 @@ private:
   // This is set after `validate()` is run at least once successfully.
   bool isValidated_ = false;
 
-  // Required workspace buffer size in bytes. Set during createVmContext()
-  // by querying the iree.abi.transients.size.constant attribute.
-  // std::nullopt indicates the graph has not been compiled yet.
-  std::optional<size_t> workspaceSize_;
+  // Bytes backing the currently loaded VMFB artifact. IREE's bytecode module
+  // may retain references to this archive, so keep it alive for as long as the
+  // VM context is alive.
+  //
+  // Keep this declared before vmContext_: members destruct in reverse
+  // declaration order, and the context must be released before the backing VMFB
+  // bytes.
+  std::vector<uint8_t> loadedArtifactBytes_;
 
   // IREE VM context lifetime managed by the `Graph` object
   // (deleted when the `Graph` object goes out of scope).
@@ -707,6 +708,11 @@ private:
   // Avoids repeated function lookup on every execute() call.
   std::optional<iree_vm_function_t> vmFunction_;
 
+  // Required workspace buffer size in bytes. Set during createVmContext()
+  // by querying the iree.abi.transients.size.constant attribute.
+  // std::nullopt indicates the graph has not been compiled yet.
+  std::optional<size_t> workspaceSize_;
+
   // Pre-computed VM input list capacity for iree_vm_list_create().
   // Set during createVmContext() to avoid recomputing on every execute().
   iree_host_size_t vmInputListCapacity_ = 0;
@@ -714,11 +720,6 @@ private:
   // Backend for the currently loaded runtime state. `execute()` requires a
   // handle for this backend because VMFB artifacts are backend-specific.
   std::optional<Backend> loadedBackend_;
-
-  // Bytes backing the currently loaded VMFB artifact. IREE's bytecode module
-  // may retain references to this archive, so keep it alive for as long as the
-  // VM context is alive.
-  std::vector<uint8_t> loadedArtifactBytes_;
 
   // Compile-side cache set by `getCompiledArtifact()`. The AOT
   // `loadFromArtifact()` API uses caller-provided VMFB bytes directly and does
--- a/include/fusilli/graph/graph.h
+++ b/include/fusilli/graph/graph.h
@@ -111,13 +111,43 @@ public:
     return ok();
   }
 
-  // Compiles the graph using IREE compiler and sets up the IREE VM
-  // context for future g->execute calls.
+  // Compiles the graph using IREE compiler and sets up the IREE VM context for
+  // future g->execute calls. This is the default JIT convenience API.
+  //
+  // This is equivalent to:
+  //   compileToArtifact(handle.getBackend(), remove)
+  //   loadFromArtifact(handle, artifactPath)
+  //
+  // The compiled artifact is backend-specific and is immediately loaded using
+  // the same backend from `handle`.
   //
   // Set `remove = true` to remove compilation artifacts (cache files) when
   // this `Graph` instance goes out of scope.
   ErrorObject compile(const Handle &handle, bool remove = false) {
     FUSILLI_LOG_LABEL_ENDL("INFO: Compiling Graph");
+    FUSILLI_ASSIGN_OR_RETURN(auto vmfbPath,
+                             compileToArtifact(handle.getBackend(), remove));
+    return loadFromArtifact(handle, vmfbPath);
+  }
+
+  // Compiles the graph using IREE compiler to produce a backend-specific VMFB
+  // artifact. This does not create any runtime state; call
+  // `loadFromArtifact()` before executing.
+  //
+  // If callers need artifacts for multiple backends, they should copy or
+  // serialize each returned VMFB path before compiling the next backend.
+  // Fusilli owns only the in-process `cache_` for the most recent compile-side
+  // artifact; it does not provide persistent artifact storage or invalidation.
+  //
+  // Compiling a new artifact does not load or unload runtime state. If this
+  // `Graph` already has an artifact loaded, that artifact remains executable
+  // until a later `loadFromArtifact()` replaces it.
+  //
+  // Set `remove = true` to remove compilation artifacts (cache files) when
+  // this `Graph` instance goes out of scope.
+  ErrorOr<std::filesystem::path> compileToArtifact(Backend backend,
+                                                   bool remove = false) {
+    FUSILLI_LOG_LABEL_ENDL("INFO: Compiling Graph to artifact");
     FUSILLI_RETURN_ERROR_IF(!isValidated_, ErrorCode::NotValidated,
                             "Graph must be validated before being compiled");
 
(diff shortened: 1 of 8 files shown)


MaheshRavishankar:
Reviewing this PR in the context of <Insert PR> within my agent, this is the outcome of my discussion with it. I am just posting it here cause it is consistent to what my understanding is.  Also below it flags the std::filesystem usage, which did stick out. I would rather not have the API depend on directly reading from the filesystem but rely on passing the artifact back in memory, with the use of filesystem being a convenience feature. Just to clarify: These are comments I'd like to document the answers for. Not necessarily holding up the PR for this.

///==--- Start of output from discussion in agent --- ==///

`loadFromArtifact` can be split, and that is probably the right shape for hipDNN serialization.

`loadFromArtifact(handle, path)` currently does too much as one operation:

  1. validates graph state
  2. clears runtime state
  3. creates the IREE VM context
  4. registers the HAL module/device group
  5. reads the VMFB from a path
  6. creates/registers the bytecode module
  7. resolves module.main or module.main$async
  8. computes VM input capacity
  9. queries workspace size

  The natural split is:

```
  graph.load(handle);          // create VM context + register HAL module
  graph.addArtifact(vmfbBytes); // create/register bytecode module + finalize runtime state
```

  or, less ambiguous:

```
  graph.createRuntimeContext(handle);
  graph.loadArtifactBytes(vmfbBytes);
```

  For the hipDNN use case, I would prefer exposing a bytes API directly, not forcing VMFB bytes through a temp file:

```
  ErrorObject loadFromArtifactBytes(const Handle& handle,
                                    std::span<const uint8_t> vmfb);
```

  Then keep the path API as a convenience wrapper:

```
  ErrorObject loadFromArtifact(const Handle& handle,
                               const std::filesystem::path& path) {
    FUSILLI_CHECK_ERROR(load(handle));
    return addArtifact(readFile(path));
  }
```

  The important detail in createVmContext is that only the first half is truly context setup: VM context creation and HAL module registration. The bytecode module load is independent enough to factor out.
  After the artifact is registered, Fusilli still must resolve the entry point and query workspace size; those cannot happen before addArtifact.

  One caveat: addArtifact needs clear failure semantics. Once a VM module is registered into an IREE VM context, there is no obvious “unregister this module and retry” path. So if addArtifact fails after
  partial registration, that should probably be an unrecoverable error.

  For hipDNN/Fusilli serialization, the flow would become:

```
  // serialize
  artifactPath = graph.compileToArtifact(handle.getBackend());
  vmfbBytes = read artifactPath;
  payload = fusilli header + graph metadata/UID metadata + vmfbBytes;

  // deserialize
  import hipDNN graph only for Fusilli metadata/UID mapping;
  graph.loadFromArtifactBytes(handle, vmfbBytes);
```

  That avoids compilation during deserialization. You still need either graph import or equivalent serialized ABI/UID metadata, because VMFB alone does not tell the hipDNN plugin how UIDs map to Fusilli tensor
  attrs / call argument order.

rsuderman:
Is there a reason we can't have a `compile-save-load` behavior. Specifically `compile` would invoke `iree-compile` and load the VMFB into the current vm context. `save` would take the compiled graph's bytes and save them to an output source (probably just a `std::vector<uint8>`). And load would parse an output `vmfb` from bytes and load it into the context.

This aligns with some of my thoughts on clearly delineating generating an executable artifact and loading it into the context.

It may also be a good time to move away from the file based interactions and instead performing compilation in memory. Materializing the intermediary files / handling everything with paths just feels like extra I/O.

MaheshRavishankar:
> Is there a reason we can't have a `compile-save-load` behavior. Specifically `compile` would invoke `iree-compile` and load the VMFB into the current vm context. `save` would take the compiled graph's bytes and save them to an output source (probably just a `std::vector<uint8>`). And load would parse an output `vmfb` from bytes and load it into the context.
> 

Why do we need the VMContext for compilation? In my mind I think the VMContext is only needed for the execution side of things?

> This aligns with some of my thoughts on clearly delineating generating an executable artifact and loading it into the context.
> 
> It may also be a good time to move away from the file based interactions and instead performing compilation in memory. Materializing the intermediary files / handling everything with paths just feels like extra I/O.

+1

rsuderman:
> Why do we need the VMContext for compilation? In my mind I think the VMContext is only needed for the execution side of things?

I actually agree that compilation and loading should be separately invoked exactly for the serialization use case. My comment was more focused on contrasting to our existing workflow.

sjain-stanford:
Thanks for taking a look. Let me address the broad topics discussed above:

- _Loading from serialized bytes:_ 
It is fairly straightforward to support loading compiled vmfb from raw `std::span<const uint8_t>` bytes instead of (or in addition to) filesystem paths - just a small convenience wrapper. For context, historically fusilli started off with CLI compile invocation so file IO was inevitable, and the cache system was built around this - [CacheFile](https://sourcegraph.com/r/github.com/iree-org/fusilli/-/blob/include/fusilli/support/cache.h?L69-71). The CAPI path carried this forward (native C API in IREE does support in-memory bytecode module). For keeping the external API surface stable, I'll switch over `compileToArtifact` and `loadFromArtifact` APIs to use raw bytes with the underlying calls still reading / writing to disk (keeping things in sync with the current state). To keep concerns separate and scope of this PR focused on enabling AOT workflow, in a follow-on I can add support for native in-memory artifacts for the C API path and use disk serialization for CLI path. This should anyways be non-blocking for the hipdnn serialization work as long as you get the byte interface on the API.

- _Why do we need the VMContext for compilation?_
We don't for the AOT / serialized API. Looking at `compileToArtifact` you'll see it no longer takes a handle/device and doesn't populate the VM context. The VM context is populated in the `loadFromArtifact` call. The JIT API does, but that's OK because compile is an implementation detail there (not intended to be run decoupled from execute).

- _Splitting `loadFromArtifact` down further_
We could consider this although I'm not sure what the benefits are / how they stack against the risks of reusing a pre-populated VM context from a stale run. Improper caching of stale state is a painful area to debug when things go wrong. Splitting also increases the error surface of having users call these low level APIs improperly so leaning towards a higher level `loadFromArtifact` is better from that standpoint.

include/fusilli/graph/graph.h
```diff
@@ -111,29 +112,92 @@ class Graph : public INode {
...
+                               std::span<const uint8_t> vmfbBytes) {
+    FUSILLI_LOG_LABEL_ENDL("INFO: Loading compiled artifact into VM context");
+    FUSILLI_RETURN_ERROR_IF(
+        !isValidated_, ErrorCode::NotValidated,
+        "Graph must be validated before loading a compiled artifact");
+    std::vector<uint8_t> artifactBytes(vmfbBytes.begin(), vmfbBytes.end());
+    // Loading replaces the currently executable artifact. Clear first so a
+    // failed load cannot leave execute() using stale runtime state from an
+    // older artifact.
+    clearRuntimeState();
```
rsuderman, line 192: All this manipulation of runtime state feels problematic. Loading an artifact into an already loaded graph feels like it should be illegal if it previously failed. Also, cleanup order should be defined by declaration order in the class.

include/fusilli/graph/graph.h
```diff
@@ -397,8 +462,19 @@ class Graph : public INode {
...
   // Definition in `fusilli/backend/runtime.h`.
-  ErrorObject createVmContext(const Handle &handle,
-                              const std::string &vmfbPath);
+  ErrorObject createVmContext(const Handle &handle);
+
+  void clearRuntimeState() {
+    // Teardown order matters: the IREE bytecode module owned by vmContext_ may
+    // retain references into loadedArtifactBytes_, so release vmContext_ before
+    // clearing the backing VMFB bytes.
+    vmContext_.reset();
```
rsuderman: See comment about declaration order.

rsuderman (changes requested):

include/fusilli/graph/graph.h
sjain-stanford: Addressed.

include/fusilli/graph/graph.h
sjain-stanford: The usecase that motivated this change (compile for N backends, then switch between them at execute time) was pretty contrived as well (would kill overall latency if we did that in a real setting) but here we are. Desperate times call for desperate measures ;) 

The options we have just boil down to:
- poison on failed load - could make a failed `loadFromArtifact` unrecoverable for that graph instance, returning an error telling users to create a new graph - this seems unnecessarily strict though and maybe surprising for users if say the failure was just invalid bytes and they wanted to retry with valid bytes
- having graph instance hold state on more than one compilation / runtime context - that would've been horrific to deal with and it breaks the abstraction of per-graph runtime state and lifetime tracking
- force new graph creation for a new load (so only allow one load clean load per graph instance) - feasible but again overly restrictive - maybe provide a user API to reset?
- allow repeat loads but auto clear state between them (even failed ones) - as long as it's clearly documented on the API, I believe this is the most reasonable contract:
```
  // A `Graph` owns one loaded runtime state at a time. Loading a new artifact
  // replaces any previously loaded VM context, function, workspace size, and VM
  // input list capacity. If loading fails, the `Graph` is left without loaded
  // runtime state. To keep multiple backends loaded concurrently, use one
  // `Graph` instance per backend artifact. To switch backends on one `Graph`,
  // call `loadFromArtifact()` with the selected backend artifact before
  // `execute()`.
```

> Loading an artifact into an already loaded graph feels like it should be illegal if it previously failed. 

If it previously failed, there wouldn't be a runtime state to begin with (we clear it). Users should expect `loadFromArtifact` to clear any previous runtime state - this aligns with the intent of the API as named - why load when you want to preserve previous state? Why preserve previous state when you're loading a newly compiled artifact? Also, there's no reason to have partially loaded runtime state, so we clear those as well.

Taking a step back, I think a more realistic usecase this enables will just be having separate graph instances load the corresponding backend-specific compiled artifacts, independent of each other so we wouldn't see much of this being an issue.

sjain-stanford: To close the loop on this: I do think what we currently have is in a good sweet spot but I'm also open to the idea of rejecting an artifact reload on an already loaded graph instance if you feel strongly about it. If that's the way we want to go, we'd need to support multiple artifact loads through either 1) separate graph instances or 2) expose user facing API to explicitly reset graph state between loads.

PR #378, opened 2026-04-27
Add TANH_BWD pointwise op

rsuderman:
Emit torch.aten.tanh_backward by first computing tanh(x), then applying the backward formula dy * (1 - tanh(x)^2). Adds emitter case, sample test, and minimal lit test.

Signed-off-by: Rob Suderman <rob.suderman@gmail.com>


sjain-stanford (approved):
LGTM!

PR #393, opened 2026-05-05
[1/N Dynamic Shapes]: Defer workspace size queries until execute for access to runtime input shapes

sjain-stanford:
For the static case, workspace size was a constant and independent of the runtime input sizes so it was OK to query and cache it during VM context creation. However for the dynamic case it is dependent on the runtime input shapes. 

This PR moves workspace-size querying out of VM context creation and into the explicit `getWorkspaceSize()` call. This keeps compile/load from querying transient size and makes `execute()` require callers to query workspace size before allocating and passing a workspace buffer.

For this PR, `getWorkspaceSize()` remains independent of runtime inputs: it queries the currently supported constant transient size, stores the maximum queried size on the graph, and returns `std::nullopt` when no runtime artifact is loaded. Variant-pack/handle-dependent dynamic workspace sizing is intentionally deferred until the dynamic size function path is added.

Updates samples, benchmarks, and tests to call `getWorkspaceSize()` before `allocateWorkspace(...)` and then pass the resulting workspace to `execute()`. Also covers the AOT compile/load path.

### DIFF
--- a/samples/pointwise/pointwise_binary_cmp_ops.cpp
+++ b/samples/pointwise/pointwise_binary_cmp_ops.cpp
@@ -80,98 +80,100 @@ TEST_CASE("Pointwise binary compare ops", "[pointwise][graph]") {
     return std::make_tuple(graph, x0T, x1T, yT);
   };
 
-  auto testDtype = [&]<typename T>(
-                       Handle &handle, DataType dt,
-                       const std::vector<std::pair<T, T>> &valuePairs) {
-    auto [graph, x0T, x1T, yT] = buildNewGraph(handle, dt);
+  auto testDtype =
+      [&]<typename T>(Handle &handle, DataType dt,
+                      const std::vector<std::pair<T, T>> &valuePairs) {
+        auto [graph, x0T, x1T, yT] = buildNewGraph(handle, dt);
 
-    for (auto [x0, x1] : valuePairs) {
-      // Allocate input buffers.
-      FUSILLI_REQUIRE_ASSIGN(auto x0Buf,
-                             allocateBufferOfType(handle, x0T, dt, x0));
-      FUSILLI_REQUIRE_ASSIGN(auto x1Buf,
-                             allocateBufferOfType(handle, x1T, dt, x1));
+        for (auto [x0, x1] : valuePairs) {
+          // Allocate input buffers.
+          FUSILLI_REQUIRE_ASSIGN(auto x0Buf,
+                                 allocateBufferOfType(handle, x0T, dt, x0));
+          FUSILLI_REQUIRE_ASSIGN(auto x1Buf,
+                                 allocateBufferOfType(handle, x1T, dt, x1));
 
-      // Allocate output buffer.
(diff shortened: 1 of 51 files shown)


samples/pointwise/pointwise_binary_cmp_ops.cpp
```diff
@@ -80,98 +80,100 @@ TEST_CASE("Pointwise binary compare ops", "[pointwise][graph]") {
     return std::make_tuple(graph, x0T, x1T, yT);
   };
 
-  auto testDtype = [&]<typename T>(
-                       Handle &handle, DataType dt,
-                       const std::vector<std::pair<T, T>> &valuePairs) {
```
sjain-stanford, line 85: This diff is spurious mostly. After we changed the workspace allocation block inside this templated lambda, clang-format made a different stable formatting choice for the enclosing lambda assignment.

sjain-stanford:
@AaronStGeorge this is first of a series of PRs to plumb support for dynamic shapes. These will be breaking changes for the plugin hence looping you in on it :) In this one `Graph::getWorkspaceSize` signature hasn't changed much (except it now propagates error to the caller) but in the next iterations it will need to take in the handle (for executing size functions) and the runtime inputs from variant pack (to get workspace size). Just FYI so you can plan on the plugin side changes accordingly.

sjain-stanford:
@BrianHarrisonAMD this is the PR we discussed, PTAL. I intentionally kept the `Graph::getWorkspaceSize` API same as before (to minimize fusilli -> plugin -> hipdnn breaking changes) but this is just the first of N PRs. In a follow-on I'll have `Graph::getWorkspaceSize` take in a handle and the variantPack as well.

BrianHarrisonAMD:
> @BrianHarrisonAMD this is the PR we discussed, PTAL. I intentionally kept the `Graph::getWorkspaceSize` API same as before (to minimize fusilli -> plugin -> hipdnn breaking changes) but this is just the first of N PRs. In a follow-on I'll have `Graph::getWorkspaceSize` take in a handle and the variantPack as well.

Sounds good!

I think for hipDNN we will do:
- Phase 1:
     - Graph has a flag marking it as dynamic.
     - Graph describes max possible shapes at compile time.
     - Workspace remains the same, and represents the max possible needed workspace for the dynamic graph.
- Phase 2:
    - Tensors have a flag for dynamic.
    - Dynamic tensors allow wild cards dims, and strides are described as an expected ordering.
    - New workspace API added to frontend that takes in the override dims / strides.
    - Workspace API without override details will error if the graph has dynamic tensors (invalid).
    - New workspace query API for providers that takes in the override details will be added.

I think that lines up with how you are adding it to Fusilli, and connecting the two should line up nicely.

Note: There will need to be a couple version bumps to the plugin API version since these include changes to the graph & tensor schema. Once it's in we can discuss the changes needed for fusilli-provider to support these new features.

sjain-stanford:
Going to go ahead and land it so I can make forward progress (hard to keep things stacked on branches), but happy to address / fix forward if there are comments.

PR #399, opened 2026-05-07
[2/N Dynamic Shapes]: Support dynamic dims for single-dispatch (zero-workspace) graphs

sjain-stanford:
This is the second dynamic-shapes slice. It enables Fusilli graphs to keep representative static extents in `TensorAttr` while emitting annotated dimensions as dynamic (`?`) in Torch MLIR types, which lets IREE compile and run zero-transient/single-dispatch dynamic-shape graphs.

This PR covers:
- per-dimension dynamic-dim metadata on `TensorAttr`
- Torch MLIR type emission for dynamic tensor and vtensor dims
- custom-op placeholder type emission with dynamic tensor dims
- end-to-end dynamic batch sample coverage for zero-workspace convolution
- an expected-failure MIN/MAX reduction sample documenting the current dynamic-workspace limitation

Scope and follow-up:
- Working path: single-dispatch / zero transient workspace cases. This has been tested with batch dynamism on convolution-style cases.
- Follow-up: dynamic transient workspace size support for multiple-dispatch cases. Those modules expose `iree.abi.transients.size` and require invoking the runtime size function before execution.

---

Co-authored-by: GPT 5.5 <codex@openai.com>

Generated with [Codex](https://openai.com/codex)

### DIFF
--- a/include/fusilli/node/custom_op_node.h
+++ b/include/fusilli/node/custom_op_node.h
@@ -89,7 +89,7 @@ public:
   //   {FUNC_NAME}                    — node name
   //   {IN<i>_DTYPE}/{OUT<i>_DTYPE}   — element type (e.g., "f32")
   //   {IN<i>_TYPE}/{OUT<i>_TYPE}     — full value tensor type
-  //   {IN<i>_DIM<j>}/{OUT<i>_DIM<j>} — representative logical dimension
+  //   {IN<i>_DIM<j>}/{OUT<i>_DIM<j>} — logical dimension, or `?` when dynamic
   //
   // Definition in asm_emitter.h (needs getTensorTypeAsm()).
   std::string resolveMlirPlaceholders() const;
--- a/include/fusilli/node/custom_op_node.h
+++ b/include/fusilli/node/custom_op_node.h
@@ -89,7 +89,7 @@ public:
   //   {FUNC_NAME}                    — node name
   //   {IN<i>_DTYPE}/{OUT<i>_DTYPE}   — element type (e.g., "f32")
   //   {IN<i>_TYPE}/{OUT<i>_TYPE}     — full value tensor type
-  //   {IN<i>_DIM<j>}/{OUT<i>_DIM<j>} — single logical dimension
+  //   {IN<i>_DIM<j>}/{OUT<i>_DIM<j>} — representative logical dimension
   //
   // Definition in asm_emitter.h (needs getTensorTypeAsm()).
   std::string resolveMlirPlaceholders() const;
--- a/include/fusilli/support/asm_emitter.h
+++ b/include/fusilli/support/asm_emitter.h
@@ -2495,7 +2495,7 @@ inline std::string CustomOpNode::resolveMlirPlaceholders() const {
     const auto &dims = inputs[i]->getDim();
     for (size_t d = 0; d < dims.size(); ++d)
       replaceAll(mlir, "{IN" + iStr + "_DIM" + std::to_string(d) + "}",
-                 std::to_string(dims[d]));
+                 inputs[i]->isDynamicDim(d) ? "?" : std::to_string(dims[d]));
   }
   for (size_t i = 0; i < outputs.size(); ++i) {
     std::string iStr = std::to_string(i);
@@ -2507,7 +2507,7 @@ inline std::string CustomOpNode::resolveMlirPlaceholders() const {
     const auto &dims = outputs[i]->getDim();
     for (size_t d = 0; d < dims.size(); ++d)
       replaceAll(mlir, "{OUT" + iStr + "_DIM" + std::to_string(d) + "}",
-                 std::to_string(dims[d]));
+                 outputs[i]->isDynamicDim(d) ? "?" : std::to_string(dims[d]));
   }
   return mlir;
 }
--- a/include/fusilli/support/asm_emitter.h
+++ b/include/fusilli/support/asm_emitter.h
@@ -48,6 +48,7 @@
 #include <cstdint>
 #include <format> // C++20
 #include <memory>
+#include <numeric>
 #include <sstream>
 #include <string>
 #include <string_view>
@@ -336,6 +337,32 @@ inline std::string getScalarItemOpsAsm(const std::string &prefix,
 //                       /*useLogicalDims=*/false)
 //        --> "!torch.tensor<[2,4,3],f32>"
 //
+// Dynamic dims are tracked by logical dimension index. For example, marking
+// logical dim 2 (sequence length) dynamic:
+//
+//    TensorAttr dynT;
+//    dynT.setName("tensor")
+//      .setDataType(DataType::Float)
+//      .setDim({2, 3, 4}) // batch, channels, sequence length
+//      .setStride({12, 1, 3})
+//      .setDynamicDims({2})
+//
+//    dynT.getTensorTypeAsm(/*isValueTensor=*/true,
+//                            /*useLogicalDims=*/true)
+//        --> "!torch.vtensor<[2,3,?],f32>"
+//
+//    dynT.getTensorTypeAsm(/*isValueTensor=*/false,
+//                            /*useLogicalDims=*/true)
+//        --> "!torch.tensor<[2,3,?],f32>"
+//
+//    dynT.getTensorTypeAsm(/*isValueTensor=*/true,
+//                            /*useLogicalDims=*/false)
+//        --> "!torch.vtensor<[2,?,3],f32>"
+//
+//    dynT.getTensorTypeAsm(/*isValueTensor=*/false,
+//                            /*useLogicalDims=*/false)
+//        --> "!torch.tensor<[2,?,3],f32>"
+//
 // Scalars (dim={1}, stride={1}) also work through this path:
 //
 //    TensorAttr s(2.0f);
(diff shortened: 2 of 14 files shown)


sjain-stanford:
Also, cc: @BrianHarrisonAMD for visibility

include/fusilli/support/asm_emitter.h
```diff
@@ -336,6 +337,32 @@ inline std::string getScalarItemOpsAsm(const std::string &prefix,
...
+//                            /*useLogicalDims=*/true)
+//        --> "!torch.tensor<[2,3,?],f32>"
+//
+//    dynT.getTensorTypeAsm(/*isValueTensor=*/true,
+//                            /*useLogicalDims=*/false)
+//        --> "!torch.vtensor<[2,?,3],f32>"
+//
+//    dynT.getTensorTypeAsm(/*isValueTensor=*/false,
+//                            /*useLogicalDims=*/false)
+//        --> "!torch.tensor<[2,?,3],f32>"
```
BrianHarrisonAMD, line 364: What does the Dim value represent when the dimension is flagged as dynamic?
I could do something similar in hipDNN instead of using -1 for wild card dimensions.

I wasn't planning on adding a separate flag to track dynamic dims, but if we are intending that the value represents the max possible, and flagging the dimension as dynamic means it can be <= the dimension value, then I could see that being better. -1 to represent dynamic felt a bit like a hack anyways, and I was on the fence about using a separate field to track it explicitly.

sjain-stanford: We take the static values for all dims even when they're marked dynamic, but then overlay it with the dynamic annotations when emitting the MLIR assembly (so it doesn't specialize for these dims). This helps with a few things but primarily it keeps the regular shape/stride inference logic happy and gives us one representative case (specialization) of the dynamic case for validation checks. No need to change how to represent strides etc. Incidentally this also aligns perfectly with cudnn which expects fully described (static) example shapes at plan build time and then fully described overridden shapes at runtime. 

Note that we only infer static shapes / strides, not dynamicity - i.e. users are expected to mark which dims in outputs are dynamic, and not rely on dynamicity to be inferred.

include/fusilli/node/custom_op_node.h
```diff
@@ -89,7 +89,7 @@ class CustomOpNode : public NodeCRTP<CustomOpNode> {
   //   {FUNC_NAME}                    — node name
   //   {IN<i>_DTYPE}/{OUT<i>_DTYPE}   — element type (e.g., "f32")
   //   {IN<i>_TYPE}/{OUT<i>_TYPE}     — full value tensor type
-  //   {IN<i>_DIM<j>}/{OUT<i>_DIM<j>} — single logical dimension
+  //   {IN<i>_DIM<j>}/{OUT<i>_DIM<j>} — representative logical dimension
```
IanWood1: `resolveMlirPlaceholders()` should probably be updated to emit `?` for dynamic dimensions, not the representative extent. Full `{IN*_TYPE}` / `{OUT*_TYPE}` placeholders are dynamic-aware now, but users can also build types from `{IN*_DIM*}` placeholders.

samples/dynamic_shapes/reduction_min_max_dynamic_batch.cpp
```diff
@@ -0,0 +1,118 @@
...
+  auto xT = graph->tensor(
+      TensorAttr().setName("x").setDim(xDims).setDynamicDims({0}).setStride(
+          generateStrideFromDim(xDims,
+                                getContiguousStrideOrder(xDims.size()))));
+
+  auto reductionAttr = ReductionAttr().setMode(mode);
+  auto yT = graph->reduction(xT, reductionAttr);
+  yT->setName("result")
+      .setDim(yDims)
+      .setDynamicDims({0})
```
IanWood1, line 46: Is it expected that users will have to do manual dynamic dim inference for output dims? Right now the operations infer output dimensions but do set any of the dims to dynamic.

include/fusilli/support/asm_emitter.h
```diff
@@ -355,10 +382,24 @@ inline std::string TensorAttr::getTensorTypeAsm(bool isValueTensor,
   oss << (isValueTensor ? "!torch.vtensor<[" : "!torch.tensor<[");
 
   std::vector<int64_t> dims = useLogicalDims ? getDim() : getPhysicalDim();
+  std::vector<int64_t> logicalDimIndices(dims.size());
+  if (useLogicalDims)
+    std::iota(logicalDimIndices.begin(), logicalDimIndices.end(), 0);
+  else
+    logicalDimIndices = getLogicalToPhysicalPermuteOrder();
+  std::vector<size_t> dimPositions(dims.size());
+  std::iota(dimPositions.begin(), dimPositions.end(), 0);
```
IanWood1, line 391: I think this relies on an important contract that should be documented. For dynamic tensors, `stride_` is not really an actual runtime stride vector. It is just a pattern used to derive a logical ↔ physical permutation (via `getLogicalToPhysicalPermuteOrder()`).

For example, with a dense row-major logical tensor `{A, B, C}`, the representative strides are `{B*C, C, 1}`. If `C` is dynamic, the runtime strides become `{B*C_runtime, C_runtime, 1}`, so both the `A` and `B` strides are dynamic even though only one dimension is marked dynamic.

This is probably okay because Fusilli’s dynamic shape support is intentionally limited to dense layout permutations due to MLIR tensor constraints. Also, representative unit dynamic dims seem risky because they are currently treated as layout-irrelevant when deriving the permutation, even though they may become non-unit at runtime.

samples/dynamic_shapes/reduction_min_max_dynamic_batch.cpp
sjain-stanford: That's correct, as I mentioned [here](https://github.com/iree-org/fusilli/pull/399#discussion_r3209842677). The dynamic dim annotations are explicit atm as we don't infer dynamicity (i.e. which of the output dims are dynamic). Shape + stride inference propagates based on the static representative values. 

There can be weird edge cases with inferring dynamicity - for example with data-dependent dynamism, like `torch.non_zero`, outputs maybe dynamic even when inputs aren't. So left it to be explicit, but for most of the ops we support I think we might be able to infer it. Shouldn't be too difficult to enable this in a restricted setting, and then rely on user-annotations for the ones that are ambiguous to infer. WDYT?

include/fusilli/support/asm_emitter.h
sjain-stanford: > For dynamic tensors, stride_ is not really an actual runtime stride vector. It is just a pattern used to derive a logical ↔ physical permutation (via getLogicalToPhysicalPermuteOrder()).

Yeah precisely! Given a shape `[B, H?, S?, D]`, only the strides whose running product touches `H` or `S` are dynamic. 

We could have an internal helper to track this (not user-facing): 
```cpp
// Index-aligned with stride_. Derived from isDynamicDim_ + permutation.
// Computed lazily or in validate(). No user setter.
std::vector<bool> isDynamicStride_;
```

Then compute in `TensorAttr::validate()` after `dim_`/`stride_`/`isDynamicDim_`
are all known:

```cpp
// For each position in physical/stride order, stride is dynamic iff any dim
// "inside" it (i.e. strictly interior in the permutation) is dynamic.
auto perm = getPhysicalToLogicalPermuteOrder(); // reuses existing helper
isDynamicStride_.assign(stride_.size(), false);
bool seenDyn = false;
for (int i = perm.size() - 1; i >= 0; --i) {
  size_t logicalIdx = perm[i];
  if (isDynamicDim(logicalIdx)) seenDyn = true;
  // The stride at this physical position is the product of all more-interior
  // dims, so it becomes dynamic as soon as any interior dim is dynamic.
  if (i < (int)perm.size() - 1) isDynamicStride_[logicalIdx] = seenDyn;
}
```

I just don't know how useful this would be. Is there a usecase this enables / unblocks that you can see?

sjain-stanford: > Also, representative unit dynamic dims seem risky because they are currently treated as layout-irrelevant when deriving the permutation, even though they may become non-unit at runtime.

Good callout. At the moment I'm rejecting dynamic dims + broadcast dims (stride = 0). We could also reject unit dynamic dims (i.e. a dim whose representative size is 1 and is marked dynamic), to prevent the ambiguity with layout permutations on dim-1 tensors.

include/fusilli/node/custom_op_node.h
sjain-stanford: Done in https://github.com/iree-org/fusilli/pull/399/commits/b05b84ef64678617bc0ff99eb9be26c8a662334b

include/fusilli/support/asm_emitter.h
sjain-stanford: Added rejects for unit representative dim marked as dynamic, and clarified documentation around stride dynamism in https://github.com/iree-org/fusilli/pull/399/commits/b05b84ef64678617bc0ff99eb9be26c8a662334b.

samples/dynamic_shapes/reduction_min_max_dynamic_batch.cpp
sjain-stanford: Draft PR for inferring dynamic dims on outputs is here: https://github.com/iree-org/fusilli/pull/416. I'm still iterating on it but linking here for completeness.

IanWood1 (approved):

PR #410, opened 2026-05-11
Fix API breakage to make porting library easier

rsuderman:
API changes broken syncing on downstream clients. Restoring old API by add a getWorkspaceOrError function to pass the value and error handling.

### DIFF
--- a/include/fusilli/backend/runtime.h
+++ b/include/fusilli/backend/runtime.h
@@ -314,7 +314,7 @@ inline ErrorObject Graph::createVmContext(const Handle &handle) {
   return ok();
 }
 
-inline ErrorOr<std::optional<size_t>> Graph::getWorkspaceSize() {
+inline ErrorOr<std::optional<size_t>> Graph::getWorkspaceSizeOrError() {
   if (vmContext_ == nullptr)
     return ok(std::optional<size_t>());
 
@@ -324,6 +324,14 @@ inline ErrorOr<std::optional<size_t>> Graph::getWorkspaceSize() {
     workspaceSize_ = workspaceSize;
 
   return ok(workspaceSize_);
+}
+
+inline std::optional<size_t> Graph::getWorkspaceSize() {
+  auto sizeOrError = getWorkspaceSizeOrError();
+  ErrorObject err = sizeOrError;
+  if (!err.isOk())
+    return std::nullopt;
+  return *sizeOrError;
 }
 
 // Queries the required transient/workspace buffer size from the compiled
@@ -404,7 +412,7 @@ Graph::execute(const Handle &handle,
   bool executeAsync = kBackendExecuteAsync.at(handle.getBackend());
   FUSILLI_RETURN_ERROR_IF(
       !workspaceSize_.has_value(), ErrorCode::InvalidArgument,
-      "Graph::execute requires getWorkspaceSize() to be called before "
+      "Graph::execute requires getWorkspaceSizeOrError() to be called before "
       "workspace allocation and execution");
 
   iree_allocator_t allocator = iree_allocator_system();
--- a/include/fusilli/graph/graph.h
+++ b/include/fusilli/graph/graph.h
@@ -394,7 +394,13 @@ public:
   // artifact is loaded, 0 if no workspace is needed, or the maximum required
   // size in bytes seen for the currently loaded runtime state. Dynamic
   // workspace sizes are not supported yet and return an error.
-  ErrorOr<std::optional<size_t>> getWorkspaceSize();
+  std::optional<size_t> getWorkspaceSize();
+
+  // Same as above but returns an error if the query fails (e.g., dynamic
+  // workspace sizes not supported, or if the VM context isn't properly
+  // initialized). This is useful for internal calls where we want to propagate
+  // the error instead of treating it as "no workspace needed".
+  ErrorOr<std::optional<size_t>> getWorkspaceSizeOrError();
 
   // ASM emitter driver method.
   //
--- a/benchmarks/driver.cpp
+++ b/benchmarks/driver.cpp
@@ -235,7 +235,7 @@ static ErrorObject benchmarkConvFprop(const ConvOptions &opts,
   }
 
   // Allocate workspace buffer if needed.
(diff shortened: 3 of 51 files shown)


AaronStGeorge (approved):
LGTM, I'd like to hear @sjain-stanford's opinion

include/fusilli/backend/runtime.h
```diff
@@ -326,6 +326,14 @@ inline ErrorOr<std::optional<size_t>> Graph::getWorkspaceSize() {
   return ok(workspaceSize_);
 }
 
+inline std::optional<size_t> Graph::getWorkspaceSize() {
+  auto sizeOrError = getWorkspaceSizeOrError();
+  ErrorObject err = sizeOrError;
+  if (!err.isOk())
+    return std::nullopt;
+  return *sizeOrError;
+}
```
sjain-stanford, line 335: Can you mark `Graph::getWorkspaceSize` as temporary with a `[[deprecated]]` notice + a TODO:
```
TODO: Remove once downstream users have migrated to the new `getWorkspaceSizeOrError`.
```

In a world where only static shapes exist, `workspaceSize_` was evaluated and cached once since it was a compile time constant for the static cases. And `Graph::getWorkspaceSize` would read and return the private size (without re-querying it at execute phase). This would be one of:
- nullopt when not set (i.e. graph isn't compiled)
- 0 when it is not required
- non-zero when required
There was no error generating path. 

With dynamic shapes, `workspaceSize_` is no longer a compile time constant, and needs runtime evaluation. Here, `getWorkspaceSize` actually calls `queryTransientSize` each time invoked. Here the contract is:
- error if queryTransientSize fails (many possible reasons)
- nullopt when VM context is not setup (e.g. if users getWorkspaceSize before calling compile)
- 0 when it is not required (for the current invocation)
- non-zero when required (for the current invocation)

This change to classify an error status as `std::nullopt` is potentially dangerous and alters the interpretation of the contract.

include/fusilli/graph/graph.h
sjain-stanford, line 291: Please update the usage example to call `getWorkspaceSizeOrError()`, since dynamic workspace queries can now fail and the example should show the error-propagating API rather than the deprecated compatibility wrapper.

include/fusilli/graph/graph.h
```diff
@@ -394,7 +394,13 @@ class Graph : public INode {
   // artifact is loaded, 0 if no workspace is needed, or the maximum required
   // size in bytes seen for the currently loaded runtime state. Dynamic
   // workspace sizes are not supported yet and return an error.
```
sjain-stanford, line 396: This still says dynamic workspace sizes "return an error" but now swallows errors and returns `std::nullopt`. Please split the comments so `getWorkspaceSize()` is explicitly documented as a deprecated, temporary compatibility wrapper that returns `std::nullopt` both when no runtime artifact is loaded and when querying the workspace size fails, while `getWorkspaceSizeOrError()` documents the error-returning behavior.

include/fusilli/graph/graph.h
```diff
@@ -394,7 +394,13 @@ class Graph : public INode {
...
-  ErrorOr<std::optional<size_t>> getWorkspaceSize();
+  std::optional<size_t> getWorkspaceSize();
+
+  // Same as above but returns an error if the query fails (e.g., dynamic
+  // workspace sizes not supported, or if the VM context isn't properly
+  // initialized). This is useful for internal calls where we want to propagate
+  // the error instead of treating it as "no workspace needed".
```
sjain-stanford, line 402: Let's reword it to not reference "above" since the above will be removed shortly.

sjain-stanford (changes requested):
It's fine this one time but we should avoid having to support too many compat APIs for every breaking change since the complexity of keeping multiple paths active through rewrites is hard to keep tight. There are more such changes coming (example `getWorkspaceSize` will need the handle and variant pack as input) and it can get ugly if we need to have v0, v1, ... of APIs.

Hopefully with @AaronStGeorge's changes to minimize the churn in percolating breaking changes (by absorbing them in the plugin) this won't be needed eventually.

include/fusilli/backend/runtime.h
AaronStGeorge: `getWorkspaceSizeOrError` could be the temporary API as well. The migration path would be

1. Plugin switches to `getWorkspaceSizeOrError` 
2. Fusilli switches `getWorkspaceSize` over to new API, the only (known) consumer now doesn't use this API so the breaking change is "silent"
3. Plugin updates to use `getWorkspaceSize` with new API
4. Fusilli removes `getWorkspaceSizeOrError`

I don't have a strong opinion which API gets deprecated, just stating an option.

sjain-stanford: Hey, if plugin can switch over to getWorkspaceSizeOrError now (item1 in your list), then why not switch it over to the error returning `getWorkspaceSize` and skip the 4 PR back-forth to migrate? The plugin is our internal glue between hipdnn and fusilli and I am not seeing a good reason to be ceremonious with the API stability.

sjain-stanford: Synced with Aaron offline. We're in a messy situation with TheRock deadlock, so I am OK landing this and cleaning up after the fact. I don't have any better ideas other than disabling the Fusilli build, which is roughly the same duration/complexity as this one.

sjain-stanford: For now let's treat `getWorkspaceSizeOrError` as the temporary API, so we'll have to migrate things over.

sjain-stanford (approved):

include/fusilli/backend/runtime.h
rsuderman: Comments should be updated.

include/fusilli/graph/graph.h
rsuderman: Donezo

include/fusilli/graph/graph.h
rsuderman: Done.

include/fusilli/graph/graph.h
rsuderman: Done.

sjain-stanford:
Did you have local changes but forgot to push (since the comments were resolved but there's no commit reflecting the changes).

PR #413, opened 2026-05-12
[WIP] [fusilli] Emit flex_attention for SDPA

keshavvinayak01:
Emit `torch.hop_flex_attention` for Fusilli SDPA cases without an explicit tensor mask or dropout. Causal SDPA attaches a generated `mask_mod_fn`, GQA sets `enable_gqa`, and explicit masks/dropout continue to use `torch.aten.scaled_dot_product_attention` until explicit mask arguments are available.

Adds `generate_stats` support by returning the flex attention logsumexp as the `STATS` output, with sample, lit, and unit coverage for the new path.


keshavvinayak01:
@sjain-stanford could you review this? The failing CI tests would work once https://github.com/iree-org/iree/pull/24426 is in.

keshavvinayak01:
Also about the mask support that we discussed on discord, I could add that in a follow up PR too.

PR #416, opened 2026-05-12
[3/N Dynamic Shapes] Support inferring dynamic dims from inputs -> outputs in inferPropertiesNode

sjain-stanford:
Adds dynamic dimension inference for inferPropertiesNode so output tensors inherit compatible dynamic annotations from their inputs without requiring callers to restate them.

This also updates dynamic shape samples and focused node/lit coverage to assert inferred output dynamic dims across broadcast, same-shape, reduction, convolution, matmul, norm, SDPA, and custom-op flows.

Co-authored-by: GPT 5.5 <codex@openai.com>

🤖 Generated with [Codex](https://openai.com/codex)

### DIFF
--- a/include/fusilli/node/custom_op_node.h
+++ b/include/fusilli/node/custom_op_node.h
@@ -102,9 +102,6 @@ public:
     for (auto &input : inputs)
       input->fillFromContext(context);
 
-    for (auto &output : outputs)
-      inferSameShapeDynamicDimsFromInputs(output, inputs);
-
     return ok();
   }
 };
--- a/include/fusilli/node/custom_op_node.h
+++ b/include/fusilli/node/custom_op_node.h
@@ -102,6 +102,9 @@ public:
     for (auto &input : inputs)
       input->fillFromContext(context);
 
+    for (auto &output : outputs)
+      inferSameShapeDynamicDimsFromInputs(output, inputs);
+
     return ok();
   }
 };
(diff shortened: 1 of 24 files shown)


include/fusilli/node/custom_op_node.h
```diff
@@ -102,6 +102,9 @@ class CustomOpNode : public NodeCRTP<CustomOpNode> {
     for (auto &input : inputs)
       input->fillFromContext(context);
 
+    for (auto &output : outputs)
+      inferSameShapeDynamicDimsFromInputs(output, inputs);
```
sjain-stanford: @IanWood1 is this safe to do for custom ops? It copies dynamic dims from an input whose representative `getDim()` exactly equals the already-specified output `getDim()`, and it skips outputs that already have dynamic dims. But even then I'm skeptical we want this here. I'd rather avoid inferring dynamic dims for custom ops and instead have users annotate, since custom ops are opaque and can be doing stuff we don't control. I doubt we infer shapes/strides for custom ops too so this seems unnecessary. Thoughts?

sjain-stanford: Removed in https://github.com/iree-org/fusilli/pull/416/commits/d8288074a0942fc70bbf9bce12aea9caf446ad23.

IanWood1: Yeah, the user has to specify this (they already have to state the output shape). 


Why is `inferSameShapeDynamicDimsFromInputs` more complicated than just copying the dynamic dims from `input` to `output`?

sjain-stanford: >  just copying the dynamic dims from input to output?

This assumes the custom operation preserves the layout of inputs (kind of like pointwise ops) and there's no reduction / permutation / expansion which is difficult to guarantee. This utility was doing a (not great) compromise of inferring dynamic dims based on a match in the representative dims between inputs and outputs.

